PR #1, opened 2017-03-07
- added code from internal repository

mobilunity-user:


package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "ravendb-node",
```
gregolsky: Let's change the name of the package to `ravendb-client`.

package.json
```diff
@@ -0,0 +1,49 @@
...
+    "url": "git@git.devs.domerp.com:frontend-projects/ravendb-node.git"
+  },
+  "keywords": [
+    "ravendb",
+    "typescript",
+    "node",
+    "nodejs",
+    "nosql"
+  ],
+  "author": "Hybernating Rhinos",
```
gregolsky: A typo here. It's *Hibernating Rhinos LTD*

package.json
```diff
@@ -0,0 +1,49 @@
...
+  "scripts": {
+    "_prepublish": "npm run bundle",
+    "quick-build": "tsc --pretty",
+    "bundle": "node scripts/build-bundle.js",
+    "pretest": "node scripts/build-tests.js",
+    "test": "mocha"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git@git.devs.domerp.com:frontend-projects/ravendb-node.git"
```
gregolsky: Please change the url to https://github.com/ravendb/ravendb-nodejs-client

package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "keywords": [
+    "ravendb",
+    "typescript",
+    "node",
+    "nodejs",
+    "nosql"
+  ],
+  "author": "Hybernating Rhinos",
+  "license": "ISC",
```
gregolsky: Please set the `license` field to [match the one set on the repository](https://github.com/ravendb/ravendb-nodejs-client/blob/v4.0/LICENSE).

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "version": "2.2",
+  "compilerOptions": {
+    "target": "es5",
```
gregolsky: target mismatch with https://github.com/ravendb/ravendb-nodejs-client/pull/1/commits/b0919fb7aa07a8f95cf7c3cdf0ebbe3684e155be#diff-601a9423323a9d2a4fd0c9b51b5ae4caR42

src/Documents/DocumentStore.ts
```diff
@@ -0,0 +1,60 @@
+import {IDocument} from './IDocument';
+import {IDocumentStore} from './IDocumentStore';
+import {IDocumentSession} from "./Session/IDocumentSession";
+import {DocumentSession} from "./Session/DocumentSession";
+import {RequestExecutor} from '../Http/RequestExecutor';
+import {DocumentConventions} from './Conventions/DocumentConventions';
+
```
gregolsky, line 7: Let's use the following style conventions for Typescript code in this project:

Use PascalCase for class and interface names.
Use PascalCase for enum values.
Use camelCase for function names.
Use camelCase for property names and local variables.
Use camelCase for argument names.
Use "_" as a prefix for `private` properties (not for `protected` ones).
Use camelCase for file / module names;

src/Documents/DocumentStore.ts
```diff
@@ -0,0 +1,60 @@
...
+export class DocumentStore implements IDocumentStore
+{
+  protected Url: string;
+  protected Database: string;
+  protected ApiKey?: string;
+  protected _RequestExecutor: RequestExecutor;
+  protected _Conventions: DocumentConventions; 
+
+  public get RequestExecutor(): RequestExecutor
```
gregolsky: Please keep one bracket opening/closing convention - opening brackets on line end and closing ones in new lines.

scripts/build-bundle.js
```diff
@@ -0,0 +1,46 @@
...
+
+        const filtered = lines
+            .filter(line => !line.startsWith('import'))
+            .map(line => line.replace(/export /, ''))
+            .join('\n');
+
+        return `/* file: ${file} */\n${filtered}`;
+    }).join('\n');
+
+fs.writeFileSync(bundle, allContents, { encoding: 'utf8', flag: 'a' });
```
gregolsky: Are we going to convert this to `gulp` tasks ?

gregolsky (changes requested):

src/Documents/Document.ts
```diff
@@ -0,0 +1,15 @@
+import {IDocument} from './IDocument';
+
+export class Document implements IDocument {
+  public entityName(): string {
+      return this.constructor.name;
+  }
+
```
ayende, line 7: Would we need serialize / deserialize? We are working on JS, after all. We can just do `JSON.serialize` and vice versa and get the same behavior, no?

gregolsky: We probably do not need a class nor interface for `Document` at all. `Raven.Client` (C#) does not have a concept of document. It just returns objects/entities. We could just use `JSON.parse()` on the document body from server's response and `JSON.serialize()` if we need a string representation.

ayende: Agree, we should be able to accept any JS object here.

gregolsky: @a-serdukov-mobilunity please adjust the code based on our above comments. If you have any questions concerning this change let's talk tomorrow.

gregolsky:
@a-serdukov-mobilunity Please fix the conflicts with LICENSE file. Thanks.

ayende:
We need to pull this already, so there is _something_ that we can start with

gregolsky:
@ayende OK, I've rebased changes and created PR #2 . We should close this one.

src/Documents/Document.ts
mobilunity-user: how i should serialize "any js object" ? i need some predefined methods like fromJSON/toJSON. 
your Python client accepts any objects, it also support linked documents where the property is a another class

also, a lot of mongo clients has own models classes with validation, relations etc (mongoose for example)
i would use IDocument/Document based on Backbone Model - they allows more features that simple js object literal

ayende: `JSON.serialize` should handle it, no?
And validation is something that you do outside of the object, using listeners, for example.

PR #4, opened 2017-04-05
Alex/develop

mobilunity-user:


ayende:
The API for the session isn't correct. You are using a session where every call to store does an immediate call to the server. You need to wait on that until you have `saveChanges`

ayende:
What is `IDocument`?

ayende:
Same behavior WRT delete / any other write operation on the session.

ayende:
There doesn't seem to be any tests around querying?

ayende:
Or creating indexes.

ayende:
You don't have an actual session (with change tracking), you are always calling the server

gregolsky:
@a-serdukov-mobilunity We need to get rid of things like `IDocument` and `DocumentKey`. It's just an object and string. Also calling it like this:

```typescript
(session.load("products/0" as DocumentKey) as Promise<IDocument>)
                .then((document: IDocument) => { ... });
```

is inconvenient I'm afraid. This one is sufficient:

```typescript
session.load("products/0").then(document => { ... });
```

PR #5, opened 2017-05-04
Alex/develop

mobilunity-user:


src/Database/Auth/ApiKeyAuthenticator.ts
```diff
@@ -69,7 +70,7 @@ export class ApiKeyAuthenticator {
 
     return tryAuthenticate()
       .then((response: IResponse) => {
-        const body = response.body;
+        const body = response.body as IHash;
```
ayende: What is this `IHash` for?

src/Database/Auth/ApiKeyAuthenticator.ts
```diff
@@ -101,14 +102,14 @@ export class ApiKeyAuthenticator {
       let publicKey: Buffer, body: IResponseBody;
 
       if (!StatusCodes.isOk(response.statusCode)
-        && !TypeUtil.isObject(body = response.body) && !body.PublicKey
+        && !TypeUtil.isObject(body = response.body as IResponseBody) && !body.PublicKey
```
ayende, line 105: What the casting to `IResponseBody` ?

src/Database/Commands/Data/PatchCommandData.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.patchIfMissing = patchIfMissing;
+    this.additionalData = additionalData;
+    this.debugMode = debugMode
+  }
+
+  public toJson(): Object {
+    let json: Object = {
+      "Key": this.key, 
+      "Method": this.method,
+      "Patch": this.scriptedPatch.toJson(),
```
ayende, line 30: Missing `PatchIfMissing`

src/Database/Commands/DeleteByIndexCommand.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  public createRequest(serverNode: ServerNode): void {
+    this.endPoint = StringUtil.format('{url}/databases/{database}', serverNode, this.params);
+    super.createRequest(serverNode);
+  }
+
+  public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | null | void {
+    const responseBody: IResponseBody = response.body;
+    const status: StatusCode = response.statusCode;
+
```
ayende: If there is an error, shouldn't we include the status code as well?

src/Database/Commands/DeleteDatabaseCommand.ts
```diff
@@ -0,0 +1,43 @@
...
+    public createRequest(serverNode: ServerNode): void {
+        let dbName: string = this.databaseId.replace('Raven/Databases/', '');
+
+        this.params = {name: dbName};
+        this.hardDelete && this.addParams({'hard-delete': 'true'});
+        this.endPoint = StringUtil.format('{url}/admin/databases', serverNode);
+    }
+
+    public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | null | void {
+        const body: IResponseBody[] = response.body as IResponseBody[];
```
ayende: What is the meaning of this array here The response doesn't return any such thing.

src/Database/Commands/IndexQueryBasedCommand.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    this.params = {
+      pageSize: query.pageSize,
+      allowStale: options.allowStale,
+      details: options.retrieveDetails
+    };
+
+    this.endPoint += StringUtil.format('/queries/{0}', this.indexName);
+    query.query && this.addParams('query', query.query);
```
ayende, line 45: There are other params (wait for non stale, sort order, etc), where are those?

src/Database/Commands/PatchByIndexCommand.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.patch = patch;
+  }
+
+  public createRequest(serverNode: ServerNode): void {
+    if (!(this.patch instanceof PatchRequest)) {
+      throw new InvalidOperationException('Patch must me instanceof PatchRequest class');
+    }
+
+    this.payload = this.patch.toJson();
+    this.endPoint = StringUtil.format('{url}/databases/{database}', serverNode,this.params);
```
ayende: This looks _very_ different that the actual C# code, which should serve as the master template. 
We want to make sure that they are close to one another to ease maintenance in the future.

src/Database/DatabaseDocument.ts
```diff
@@ -0,0 +1,32 @@
+import {IJsonSerializable} from "../Json/IJsonSerializable";
+import {IHash} from "../Utility/Hash";
+
+export class DatabaseDocument implements IJsonSerializable {
+  protected secureSettings: {};
```
ayende: can drop this.

src/Documents/Session/IDocumentQuery.ts
```diff
@@ -1,31 +1,41 @@
...
-  get(callback?: QueryResultsCallback<DocumentQueryResult<IDocument>>): Promise<DocumentQueryResult<IDocument>>;
+export interface IDocumentQueryOptions {
+  usingDefaultOperator?: QueryOperator;
+  waitForNonStaleResults?: boolean;
+  includes?: string[];
+  nestedObjectTypes?: INestedObjectTypes;
+  withStatistics?: boolean;
+}
+
+export interface IDocumentQuery<T> {
```
ayende: Do we have a way to pass just a raw query to Raven?

src/Utility/StringUtil.ts
```diff
@@ -60,4 +63,22 @@ export class StringUtil {
...
+    if (TypeUtil.isNone(dbName) || !dbName) {
+      throw new InvalidOperationException('Empty name is not valid');
+    }
+
+    if (!/^[A-Za-z0-9_\-\.]+$/.test(dbName)) {
+      throw new InvalidOperationException('Database name can only contain only A-Z, a-z, \"_\", \".\" or \"-\"');
+    }
+  }
+
+  public static ucFirst(string: string): string {
```
ayende: Use full method names. Without reading the code, there is no way to translate `uc` to upperCase`

src/Utility/StringUtil.ts
```diff
@@ -60,4 +63,22 @@ export class StringUtil {
...
+    return string.charAt(0).toUpperCase() + string.substring(1);
+  }
+
+  public static lcFirst(string: string): string {
```
ayende: Same regarding methods names

test/session_tests/DocumentDeleteTest.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import {expect} from 'chai';
+import * as Promise from 'bluebird';
+import {IHash} from "../../src/Utility/Hash";
+import {DocumentStore} from "../../src/Documents/DocumentStore";
+import {IDocumentStore} from "../../src/Documents/IDocumentStore";
+import {string, Object} from "../../src/Documents/Object";
+import {IDocumentSession} from "../../src/Documents/Session/IDocumentSession";
+
+describe('Document delete test', () => {
```
ayende: Please also include some tests using async/await syntax for basic CRUD, to verify that it works and give people some sample code to look at.

ayende (changes requested):

src/Database/Auth/ApiKeyAuthenticator.ts
gregolsky: https://github.com/ravendb/ravendb-nodejs-client/blob/v4.0/src/Utility/Hash.ts

+1 and how is it different from `Object`?

src/Database/Commands/Data/PatchCommandData.ts
gregolsky: @ayende It's being added few lines below, if defined using `_.assign()`.

src/Database/Commands/DeleteDatabaseCommand.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        if (StatusCodes.isOk(response.statusCode)) {
+            const firstResult = _.first(body);
+
+            if(!firstResult.Deleted) {
+                throw new ErrorResponseException(firstResult.Error);
+            }
+        }
+
+        return null;
```
gregolsky: For `public setResponse(response: IResponse)` I saw we sometimes return `null` and sometimes it's just implicit `undefined` (void). Is there a difference and a reason for it ?

```
IRavenResponse | IRavenResponse[] | null | void 
```

Could it be just one of the two null or void?

See https://github.com/mobilunity-user/ravendb-nodejs-client/blob/e73c416d6f6953d3838edac3bdedd512650fe986/src/Database/Commands/CreateDatabaseCommand.ts

src/Database/Commands/GetIndexCommand.ts
```diff
@@ -0,0 +1,33 @@
...
+    this.params = {};
+    this.endPoint = StringUtil.format('{url}/databases/{database}/indexes', serverNode);
+    this.indexName && this.addParams('name', this.indexName);
+  }
+
+  public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | null | void {
+    const responseBody: IResponseBody = response.body;
+
+    if (!responseBody) {
+      return null
```
gregolsky: missing semicolon

src/Database/Commands/PatchCommand.ts
```diff
@@ -0,0 +1,64 @@
...
+export class PatchCommand extends RavenCommand {
+  protected key?: string;
+  protected patch: PatchRequest;
+  protected etag?: number = null;
+  protected patchIfMissing?: PatchRequest = null;
+  protected skipPatchIfEtagMismatch: boolean = false;
+  protected returnDebugInformation: boolean;
+  protected path: string;
+
+  constructor(key: string, patch, etag?, patchIfMissing?: PatchRequest, skipPatchIfEtagMismatch: boolean = false, returnDebugInformation: boolean = false) {
```
gregolsky: There's too many arguments. Let's wrap it in *opts* arg.

src/Database/RavenCommand.ts
```diff
@@ -86,7 +86,7 @@ export abstract class RavenCommand {
     return options;
   }
 
-  public setResponse(response: IResponse): IRavenCommandResponse | null | void {
+  public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | null | void {
```
gregolsky: For the sake of `BatchCommand` we're changing `RavenCommand` to accept object or array of objects. In C# RavenCommand is generic and BatchCommand extends it like this:

```
    public class BatchCommand : RavenCommand<BlittableArrayResult>, IDisposable
    {
...
```

Could we do it like this? `IRavenResponse | IRavenResponse[] | null | void` as a return type is really confusing. It would be great if we could just leave IRavenResponse.

src/Documents/Session/DocumentSession.ts
```diff
@@ -109,43 +116,57 @@ export class DocumentSession implements IDocumentSession {
...
           });
       })
       .catch((error: RavenException) => PromiseResolver.reject(error, null, callback));
   }
 
-  public query(documentType?: IDocumentType, indexName?: string, usingDefaultOperator?: boolean, waitForNonStaleResults: boolean = false,
-     includes?: string[], withStatistics: boolean = false
-  ): IDocumentQuery {
-    return new DocumentQuery(this, this.requestsExecutor, documentType);
+  public query<T extends Object>(documentTypeOrObjectType?: string | DocumentConstructor<T>, indexName?: string, options?: IDocumentQueryOptions): IDocumentQuery<T> {
```
gregolsky: Since all parameters are optional, perhaps we should wrap everything in opts. I saw in test a load of calls like `session.query(null, null, opts)` and this is not very convenient. Thoughts?

src/Documents/Session/DocumentSession.ts
```diff
@@ -109,43 +116,57 @@ export class DocumentSession implements IDocumentSession {
...
+
+    if (options) {
+      usingDefaultOperator = options.usingDefaultOperator || null;
+      waitForNonStaleResults = options.waitForNonStaleResults || null;
+      includes = options.includes || null;
+      withStatistics = options.withStatistics || null;
+      nestedObjectTypes = options.nestedObjectTypes || {};
+    }
+
+    return new DocumentQuery(this, this.requestsExecutor, documentTypeOrObjectType, indexName,
```
gregolsky: `INestedObjectTypes`? `DocumentConstructor<T>`? I thought we were not suppose to do full-blown entity mapping using constructors and all that. @ml054 was only asking for simple casting the result/results of `load<T>()` and `query<T>()` to T after it's parsed, not a class object construction. Thoughts?

cc @ayende

src/Json/Serializer.ts
```diff
@@ -1,50 +1,55 @@
 import {TypeUtil} from "../Utility/TypeUtil";
 import {DateUtil} from "../Utility/DateUtil";
 import {ArrayUtil} from "../Utility/ArrayUtil";
+import {INestedObjectTypes, DocumentConstructor} from "../Documents/Conventions/DocumentConventions";
 
 export class Serializer {
-  public static fromJSON<T extends Object>(className: { new(): T; }, source: Object | string, metadata: Object = {}, target?: T)
+  public static fromJSON<T extends Object>(target: T, source: Object | string, metadata: Object = {}, nestedObjectTypes: INestedObjectTypes = {}): T
```
gregolsky: If we're gonna do it this way - construct objects from results using constructors etc. , which I am not convinced we should - this class should have a solid set of tests.

test/session_tests/DocumentDeleteTest.ts
```diff
@@ -0,0 +1,66 @@
+/// <reference path="../../node_modules/@types/mocha/index.d.ts" />
+/// <reference path="../../node_modules/@types/chai/index.d.ts" />
+
+import {expect} from 'chai';
+import * as Promise from 'bluebird';
+import {IHash} from "../../src/Utility/Hash";
+import {DocumentStore} from "../../src/Documents/DocumentStore";
+import {IDocumentStore} from "../../src/Documents/IDocumentStore";
+import {string, Object} from "../../src/Documents/Object";
```
gregolsky: Here I get:
```
file: 'file:///c%3A/work/ravendb-nodejs-client/test/session_tests/DocumentDeleteTest.ts'
severity: 'Error'
message: 'Cannot find module '../../src/Documents/Object'.'
at: '9,30'
source: 'ts'
```

gregolsky: Actually `build:tests` does not build for me at all:

```

PS C:\work\ravendb-nodejs-client> gulp build:tests
[10:52:09] Using gulpfile C:\work\ravendb-nodejs-client\gulpfile.js
[10:52:09] Starting 'clean'...
[10:52:09] Finished 'clean' after 6.52 ms
[10:52:09] Starting 'build:tests'...
test\raven_command_tests\QueryCommandTest.ts(18,14): error TS2314: Generic type 'DocumentQuery<T>' requires 1 type argument(s).
test\session_tests\DocumentDeleteTest.ts(9,30): error TS2307: Cannot find module '../../src/Documents/Object'.
test\session_tests\DocumentFullTextSearchTest.ts(18,14): error TS2314: Generic type 'IDocumentQuery<T>' requires 1 type argument(s).
test\session_tests\DocumentFullTextSearchTest.ts(84,29): error TS2339: Property 'title' does not exist on type '{}'.
test\session_tests\DocumentFullTextSearchTest.ts(85,29): error TS2339: Property 'title' does not exist on type '{}'.
test\session_tests\DocumentFullTextSearchTest.ts(109,29): error TS2339: Property 'title' does not exist on type '{}'.
test\session_tests\DocumentFullTextSearchTest.ts(110,29): error TS2339: Property 'title' does not exist on type '{}'.
test\session_tests\DocumentFullTextSearchTest.ts(111,29): error TS2339: Property 'title' does not exist on type '{}'.
test\session_tests\DocumentLoadTest.ts(45,27): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentLoadTest.ts(55,27): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentLoadTest.ts(78,24): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentLoadTest.ts(100,27): error TS2339: Property 'product' does not exist on type 'Object'.
test\session_tests\DocumentLoadTest.ts(101,27): error TS2339: Property 'product' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(20,14): error TS2314: Generic type 'IDocumentQuery<T>' requires 1 type argument(s).
test\session_tests\DocumentQueryTest.ts(78,27): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(109,29): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(117,29): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(166,29): error TS2339: Property 'order' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(174,29): error TS2339: Property 'order' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(182,53): error TS2339: Property 'order' does not exist on type 'Object'.
test\session_tests\DocumentQueryTest.ts(202,31): error TS2339: Property 'product' does not exist on type 'Object'.
test\session_tests\DocumentStoreTest.ts(37,27): error TS2339: Property 'name' does not exist on type 'Object'.
test\session_tests\DocumentStoreTest.ts(46,27): error TS2339: Property 'key' does not exist on type 'Object'.
[10:52:12] TypeScript: 23 semantic errors
[10:52:12] TypeScript: emit succeeded (with errors)
[10:52:13] Finished 'build:tests' after 3.54 s
PS C:\work\ravendb-nodejs-client>
```

test/session_tests/DocumentQueryTest.ts
```diff
@@ -0,0 +1,219 @@
...
+        .then(() => done());
+      });
+  });
+
+  describe('Index checking', () => {
+    it('should equal dynamic index', (done: MochaDone) => {
+      session = store.openSession();
+      session.query().whereEquals('name', 'test101').get().then((results: Object[]) => {
+        expect(results[0].name).to.equals('test101');
+        done();
```
gregolsky: Just a note I believe you can return a Promise in `it()` callback and mocha is going to handle that for you without the need to call `done()`.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -0,0 +1,219 @@
...
+        session.query('company').whereEquals('name', 'withNesting').get()
+          .then((results: Object[]) => {
+            expect(results[0].product).to.equals('product');
+            expect(results[0]).to.equals('company')
+            done();
+          })
+      });
+    });
+
+    it('should make query with fetch terms', (done: MochaDone) => {
```
gregolsky, line 231: done() is not being called here.

gregolsky (changes requested):

test/session_tests/DocumentDeleteTest.ts
ayende: Another issue, why do we _have_ an `Object` class defined?

src/Database/Auth/ApiKeyAuthenticator.ts
mobilunity-user: just for internal use. object that can contain any properties.
if use Object type, we will have compiling errors trying to access properties (because any object property should be defined in ts).

src/Database/Auth/ApiKeyAuthenticator.ts
mobilunity-user: body can be object or array of objects. i cast to specific response type used in this request
also it just or internal use, no additional casts will be needed for library's users

src/Database/Commands/DeleteByIndexCommand.ts
mobilunity-user: status code is HTTP status code (from http request to raven db server)
it doesn't includes to responses in python client, it's used only to determine kind of error and for to throw corresponding exception

src/Database/Commands/IndexQueryBasedCommand.ts
mobilunity-user: pls, ask Idan - i just ported those params from python client

src/Database/Commands/PatchByIndexCommand.ts
mobilunity-user: the same thing - it was ported from python client

src/Documents/Session/IDocumentQuery.ts
mobilunity-user: at this moment - no (and in python client too)

src/Utility/StringUtil.ts
mobilunity-user: sorry, i took this function name from PHP)))

test/session_tests/DocumentDeleteTest.ts
mobilunity-user: probably it lefted here after global search/replace of IDocument  etc to Object/strings
i'll remove it

mobilunity-user: Yes async/await tests will be included when i add support to it.

src/Database/Commands/PatchCommand.ts
mobilunity-user: ok, wrapping all after patch?

src/Database/RavenCommand.ts
mobilunity-user: i'll research this issue
i tries to define somethign like type RavenResponse = IRavenResponse | IRavenResponse[] | null | void; but in use it requires a lot of "as" casts

src/Documents/Session/DocumentSession.ts
mobilunity-user: ok

src/Documents/Session/DocumentSession.ts
mobilunity-user: it is optional, if you want - you can pass nestedObjectTypes hash which maps class constructor to nested object property name.
if you dont wont it - just dont pass it and all nested props will be plain Object

The same with documentTypeOrObjectType. if you pass a string 'Product' you will have plain object casted to Product. if you pass Product class constructor - you will have Product class instance

src/Json/Serializer.ts
mobilunity-user: yes, it has a part of tests, and right, it's needed to add tests for object / nested objects construction

test/session_tests/DocumentDeleteTest.ts
mobilunity-user: i'll fix it. not removed all former IDocument / DocumentKey imports and i'll check those errors

src/Database/Auth/ApiKeyAuthenticator.ts
ayende: How is this different from any JS object?

src/Database/Auth/ApiKeyAuthenticator.ts
ayende: A response from ravendb must always be an object, there shouldn't be anything that returns arrays.

src/Database/Commands/DeleteByIndexCommand.ts
ayende: We do want to include that in the error, that can be important for users.

src/Database/Commands/PatchByIndexCommand.ts
ayende: @IdanHaim Please take a look here.

src/Database/Commands/IndexQueryBasedCommand.ts
IdanHaim: You can see in the session_query.py 
the non stale option is in the __call__ function 

```
def __call__(self, object_type=None, index_name=None, using_default_operator=None,
                 wait_for_non_stale_results=False, includes=None, with_statistics=False, nested_object_types=None):
```
wait_for_non_stale_results (see test_query.py)
Other params can be found in the __call__ or in the function below

src/Database/Auth/ApiKeyAuthenticator.ts
mobilunity-user: no difference. in pure js it will be plain object, it just a 'syntax sugar'.
IHash added just for do not have compiler error.
i created it as interface declaration just because cast 'as {[property: string]: any}' is too long and it's used in a few places

src/Database/Auth/ApiKeyAuthenticator.ts
mobilunity-user: no, in some cases response is array of objects (for ex DeleteIndex/DatabaseCommand) - i saw it in python client code..

src/Database/Commands/DeleteDatabaseCommand.ts
mobilunity-user: in python client this response contains in body array of objects:

`response = response.json()
            if not response[0]["Deleted"]:
                raise Exception(response[0]["Reason"])` 

so i cast it as array

src/Database/Commands/PatchByIndexCommand.ts
IdanHaim: Can you please look on class `PatchByIndexCommand(RavenCommand)` in the python
try to read the summary there.
look on this function in utils.py  `def build_path(index_name, query, options)`,
there you can notice how we really build the url.
The params is the options.
Notice the call to the method `Utils.build_path(self.index_name, self.query_to_update, self.options))` when I build the url in the create_request

src/Database/Commands/DeleteDatabaseCommand.ts
IdanHaim: I change it after see this commit ravendb/RavenDB-Python-Client#17

IdanHaim: I change it after see this commit ravendb/RavenDB-Python-Client#17

src/Database/Auth/ApiKeyAuthenticator.ts
gregolsky: @mobilunity-user But when using TS typings users would need to deal with `IHash` alias:
- see IHash, be like "what's that"
- go to definition
- "Oh it's just an object with any properties, nevermind"
- pass object
rather than go ahead and pass `object` in the first place.

src/Database/Auth/ApiKeyAuthenticator.ts
gregolsky: It's not the case in C#:

https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Operations/Indexes/DeleteIndexOperation.cs

https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Server/Documents/Handlers/IndexHandler.cs#L393

@mobilunity-user In general server should always return an object. If you spot a place, where it's not the case, please just let us know.

@IdanHaim Would you check DeleteIndex/DatabaseCommand in Python client please?

gregolsky:
I'll just paste the tickets, so that we know what changes we're looking at:
http://issues.hibernatingrhinos.com/issue/RavenDB-6736
http://issues.hibernatingrhinos.com/issue/RavenDB-6745

src/Database/Auth/ApiKeyAuthenticator.ts
ml054: Something like this: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type

mobilunity-user: object isn't equivalent to 'object with any properties': http://take.ms/TNvxz
for declare  'object with any properties' i should explicitly declare it like {[property: string]: any}
as i meet with this type in many places i created IHash interface for don't declare {[property: string]: any} everywhere

gregolsky: well, since we cannot use `object` (as it's yelling about missing properties) and I would really not like to use `any` then `IHash` aka ` {[property: string]: any}` is our best shot. However I think we need a *better name* for it. 

Just thinking out loud. It's like the old JObject thing we use to have in 3.5. Also it's made from JSON responses. How about "JsonObject" or sth similar? Thoughts?

@ayende @ml054 ?

mobilunity-user:
```
class Product {
  constructor(
    protected name: string
  ) {}
}

interface IRavenObject<T = any> {
  [propery: string]: T;
}

class App {
  public static main(argv?: string[], argc: number = 0): void {
    let plainObject: IRavenObject = {
      someProp: 'some value'
    };

    let products: Array<Product> = [
      new Product('product #1'),
      new Product('product #2')
    ];

    let productsByName: IRavenObject<Product> = {
      'product #1': new Product('product #1'),
      'product #2': new Product('product #2')
    };

    for (let i: number = 3; i <= 10; i++) {
      const key: string = `product #${i}`;
      const product: Product = new Product(key);

      products.push(product);
      productsByName[key] = product;
    }

    plainObject.anotherProp = 'anotherValue';

    console.log(plainObject);
    console.log(products);
    console.log(productsByName);
  }
}

App.main();
```

gregolsky: Additionally since we have `load()` defined as `public load<T extends Object>` and if we're not casting (want to have dynamic object we're gonna need to cast  to access any property (otherwise it will compain about missing props again, am I right @a-serdukov-mobilunity?

can we define load like `load<T = IRavenObject>(...)`?

src/Database/Commands/PatchByIndexCommand.ts
mobilunity-user: buildPath implemented in IndexQueryBasedCommand (abstract class which is base for PatchByIndexCommand and DeleteByIndexCommand)

src/Database/Auth/ApiKeyAuthenticator.ts
mobilunity-user: resolved

src/Database/Commands/DeleteDatabaseCommand.ts
mobilunity-user: resolved

src/Database/DatabaseDocument.ts
mobilunity-user: DatabaseDocument is used in CreateDatabaseCommand, i can't drop it

src/Utility/StringUtil.ts
mobilunity-user: resolved

src/Utility/StringUtil.ts
mobilunity-user: resolved

src/Documents/Session/DocumentSession.ts
mobilunity-user: resolved

src/Database/Commands/PatchCommand.ts
mobilunity-user: resolved

src/Database/Commands/DeleteDatabaseCommand.ts
mobilunity-user: resolve (void if empty response)

src/Database/Auth/ApiKeyAuthenticator.ts
ml054: About IHash:
1) If x is of type object, then:
a) ts compiler complains if we write: x.someProperty, but it doesn't complain when we write: x['someProperty'].
2) I think we both don't like second notation. 
3) Casting to any is also bad.
4) If we have type *object* it means it can be almost everything as we can't assume anything about its structure. 
5) What if you actually cast response body to some interface which depends on usage? You actually assuming that response should have some properties, so let's cast them. We use this technique when getting ajax responses in studio (for example each query response should have results + total count). If you can't cast to some well defined type, it means you can't assert given type at this point. As result it makes sense that TS complains about any property access on *object* type.
6) If possible don't use *Object*. Instead use *object*. See: http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

PR #7, opened 2017-05-18
RavenDB-7056

mobilunity-user:


ayende:
The default URL for the tests shouldn't have fiddler on it.

More importantly, the errors that the tests produce are poor, I'm getting failing tests, but no proper errors on why they failed.

PR #9, opened 2017-06-01
Raven db 6735

mobilunity-user:


src/Documents/DocumentStore.ts
```diff
@@ -85,14 +85,42 @@ export class DocumentStore implements IDocumentStore {
       .then((): IDocumentStore => this);
   }
 
-  public openSession(database?: string, forceReadFromMaster: boolean = false): IDocumentSession {
+  public async openSession(transaction: ITransaction) : Promise<void>;
+  public async openSession(options: ISessionOptions, transaction: ITransaction) : Promise<void>;
+  public openSession(database?: string, forceReadFromMaster?: boolean) : IDocumentSession;
+  public openSession(firstArg?: string | ITransaction | ISessionOptions, secondArg?: ITransaction | boolean): IDocumentSession | Promise<void> {
```
ayende: What is this `ITransaction` thingie here?

tryouts/transaction.ts
```diff
@@ -0,0 +1,102 @@
...
+    doc1 = await session.store(doc1);
+    doc2 = await session.store(doc2);    
+
+    await session.saveChanges();
+    await store.finalize();
+
+    console.log('doc1', doc1);
+    console.log('doc2', doc2);
+
+    await store.openSession(async (session: Session): Promise<void> => {
```
ayende, line 89: Nope!
We have a very well defined semantic for transactions, everything up to the `saveChanges`. We don't need this strange idea of a callback that will automatically call it.

ayende (changes requested):

tryouts/transaction.ts
mobilunity-user: @ayende but it's very useful. In many libraries / frameworks transactions are wrapped into callback and they automatically commits if no exceptions were thrown. For example in Laravel:

```
DB::transaction(function () {  
    DB::table('users')->update(['votes' => 1]);

    DB::table('posts')->delete();

    if ($somethingHappened) {
      throw new Exception('Something happened, I won't commit changes and i'll roll them back');
   }
});

```

If we do like you want, people which are accustomed to this workflow will be confused why changes won't apply.

probably we may create separate methods for each case (just open, open & do callback, open & do callback & apply) ?

src/Documents/DocumentStore.ts
mobilunity-user: @ayende type definition for callback function.
openSession callback should receive session instance and return promise

gregolsky:
I'll take a look at this one, after we're out with beta.

src/Documents/DocumentStore.ts
gregolsky: @mobilunity-user OK, I got it - we're going to use it due to lack of `using()` clause in JS. Let's just indicate it's a callback in the name then. How about renaming this to `ISessionCallback`, `ISessionActionsCallback` or  `ITransactionCallback`?

tryouts/transaction.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    session.initialize();
+    doc1 = session.create({id:1});
+    doc2 = await session.load('2');
+
+    doc1 = await session.store(doc1);
+    doc2 = await session.store(doc2);    
+
+    await session.saveChanges();
+    await store.finalize();
```
gregolsky, line 84: I presume this one's like the `Dispose()` method right and we need to call it just once on shutdown of our app?

tryouts/transaction.ts
gregolsky: > people which are accustomed to this workflow will be confused why changes won't apply.

I think they'll learn soon enough ;)

Coming from C# client I thought it's made this way to make up for the lack of `using()` clause in JS, but since there's nothing to dispose of, then we probably can avoid that. Also I think we don't want to be like others, but be consistent across different ecosystems.

On the other hand I can see the convenience of this. And there's another role of `using()` (grouping/closing a bunch of calls in some context). Since we don't have a `using()` to group our session calls for clearness sake it might have sense to introduce this callback accepting method in this client.

To sum up. I'd say we keep the callback version, **but** don't do `saveChanges()` automatically. This way it's just grouping the session calls, but that's OK. It's clear what's done with the session and when it's usage ends. Like this:

```javascript
await store.openSession(async (session: Session): Promise<void> => {
      doc1 = session.create({id:1});
      doc2 = await session.load('2');

      doc1 = await session.store(doc1);
      doc2 = await session.store(doc2);
      await session.saveChanges();
    });
```

Thoughts guys?

tryouts/transaction.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+class Runner
+{
+  public static async run(): Promise<void> {
+    const store = new Store();
+    const session = store.openSession();
+    let doc1: IObject, doc2: IObject;
+
+    session.initialize();
```
gregolsky, line 76: It's the `DocumentStore` instance that needs to be initialized, not the session.

tryouts/transaction.ts
```diff
@@ -0,0 +1,102 @@
...
+    doc1 = session.create({id:1});
+    doc2 = await session.load('2');
+
+    doc1 = await session.store(doc1);
```
gregolsky, line 80: Can't we just `session.store({id:1})` ?

gregolsky (changes requested):

ayende:
If they want to use group, they can just use:

```
{
     var session = store.openSession();
      doc1 = session.create({id:1});
      doc2 = await session.load('2');

      doc1 = await session.store(doc1);
      doc2 = await session.store(doc2);
      await session.saveChanges();
}
```

Why use a callback for that?

And I don't see why we would need the `session.create` for anything?

tryouts/transaction.ts
mobilunity-user: @gregolsky it's just a test app with "dummy" classes, not the library code
i just collected here all examples of api interfaces which were sent in github/youtrack comments + come of my experiments
you can just ignore all from /tryouts when doing code reviews - on relese it will be removed

src/Documents/DocumentStore.ts
mobilunity-user: @gregolsky ok, as we will not call saveChanges() automatically, it not will be 'transaction' in general case, so ISessionCallback will be preferred

tryouts/transaction.ts
mobilunity-user: @gregolsky ok, but how about return_unused_range ? it would be called manual aswell ?
 
```
await store.openSession(async (session: Session): Promise<void> => {
      ...
      await session.saveChanges();
    });

await store.finalize();

```

i have just an idea - if we called saveChanges() in callback and documents with ids generated via HiLo were stored - to call return_unused_range automatically
what you think about it ?

tryouts/transaction.ts
mobilunity-user: @gregolsky pls see comment above

tryouts/transaction.ts
mobilunity-user: @gregolsky pls see comment above

ayende:
No, we won't be using that at all.
I haven't seen anything to convince me that this adds anything at all.

*Hibernating Rhinos Ltd  *

Oren Eini* l CEO l *Mobile: + 972-52-548-6969

Office: +972-4-622-7811 *l *Fax: +972-153-4-622-7811

src/Documents/DocumentStore.ts
ayende: No, we'll drop this idea completely.

mobilunity-user: @ayende probably i can leave it just as optional feature ? now you can use in both ways
i just updated all unit test for use  load/store wrapped in callbacks - if we remove callback - i need to rewrite all them again (
also, anyway we  need to dispose used Hilo ranges. in Python it's done via `__exit__()` method which calls after end of `with(store) ...` block. In js there are `with` operator, but, no specific "magic" methods for it and also it's obsolette and non-reccomended for use

ayende: Optional features are a mess. And I would rather not start a new client with already obselete features.

Returning the hilo is optional, so we can skip it if needed but I agree that this is a good idea.
However, this is on the _store_ level, which isn't typically used with a `with` / `using`.

Typically you create it once and use if for the lifetime of your app

PR #10, opened 2017-06-13
RavenDB #6735 RavenDB #6739

mobilunity-user:


tryouts/transaction.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    session.initialize();
+    doc1 = session.create({id:1});
+    doc2 = await session.load('2');
+
+    doc1 = await session.store(doc1);
+    doc2 = await session.store(doc2);    
+
+    await session.saveChanges();
+    await store.finalize();
```
ayende, line 84: Why do we need this?

mobilunity-user: @ayende 

it's just a test app with "dummy" classes, not the library code
i just collected here all examples/propositions of api interfaces which were sent in github/youtrack comments + some of my experiments
you can just ignore all from /tryouts when doing code reviews - on release it will be removed

ayende: I mean, why the call to `finalize` ?

mobilunity-user: @ayende in my dummy app "finalize" means "return unused range"

gregolsky:
I don't think we should keep this kind of code in the repo. It may be
wrongly treated as some kind of example and misused by users.

wt., 13.06.2017, 18:23 użytkownik Mobilunity <notifications@github.com>
napisał:

tryouts/transaction.ts
ayende: Huh? Why would it _do_ that? 
Leaving aside @gregolsky point about needing to remove this code for fear of people assuming it an example. A hilo range does NOT live in the context of a session, it lives in the context of the _store_.

PR #13, opened 2017-06-20
RavenDB #7111, RavenDB #7112, RavenDB #7484

mobilunity-user:


gregolsky:
@a-serdukov-mobilunity Please try make PRs per ticket if possible. It's hard to grasp this amount of changes.

src/Database/Commands/GetStatisticsCommand.ts
```diff
@@ -16,10 +17,12 @@ export class GetStatisticsCommand extends RavenCommand {
...
-    const responseBody: IResponseBody = response.body;
+    const result: IRavenResponse = <IRavenResponse>super.setResponse(response);
 
-    if (responseBody && StatusCodes.isOk(response.statusCode)) {
-      return responseBody;
+    if (response.body) {
+      return result;
     }
+
+    throw new ErrorResponseException('Invalid server response');
```
gregolsky, line 26: It would be nice to include the response (body, status code...) in such exception.

src/Http/Request/RequestsExecutor.ts
```diff
@@ -162,123 +168,22 @@ export class RequestsExecutor {
...
-        }
-
-        throw new RequestException(
-          'Tried all nodes in the cluster but failed getting a response'
-        );
-      default:
-        throw new InvalidOperationException(
-          StringUtil.format('Invalid read behavior value: {readBehavior}', topology)
-        );
+        nonFailed && (nonFailedNode = node);
```
gregolsky, line 178: Using `some()` and having side-effects like this in the callback function makes code a bit cryptic, don't you think? In general I never use not *pure* functions for callbacks like this. A regular for loop would be more readable here.

gregolsky:
@a-serdukov-mobilunity I get the following error building tests:
```
PS C:\work\ravendb-nodejs-client> gulp build:tests
[10:42:08] Using gulpfile C:\work\ravendb-nodejs-client\gulpfile.js
[10:42:08] Starting 'clean'...
[10:42:08] Finished 'clean' after 41 ms
[10:42:08] Starting 'build:tests:args'...
[10:42:08] Finished 'build:tests:args' after 24 ms
[10:42:08] Starting 'build:tests'...
src\Documents\Session\DocumentSession.ts(129,11): error TS2322: Type 'Bluebird<T[]>' is not assignable to type 'void | PromiseLike<void>'.
  Type 'Bluebird<T[]>' is not assignable to type 'PromiseLike<void>'.
    Types of property 'then' are incompatible.
      Type '{ <U>(onFulfill?: (value: T[]) => U | PromiseLike<U>, onReject?: (error: any) => U | PromiseLike<...' is not assignable to type '<TResult1 = void, TResult2 = never>(onfulfilled?: (value: void) => TResult1 | PromiseLike<TResult...'.
        Types of parameters 'onFulfill' and 'onfulfilled' are incompatible.
          Type '(value: void) => any' is not assignable to type '(value: T[]) => any'.
            Types of parameters 'value' and 'value' are incompatible.
              Type 'T[]' is not assignable to type 'void'.
[10:42:12] TypeScript: 1 semantic error
[10:42:12] TypeScript: emit succeeded (with errors)
[10:42:12] Finished 'build:tests' after 3.7 s
PS C:\work\ravendb-nodejs-client>
```

src/Database/Commands/PatchByIndexCommand.ts
```diff
@@ -29,17 +29,16 @@ export class PatchByIndexCommand extends IndexQueryBasedCommand {
...
+    const result: IRavenResponse = <IRavenResponse>super.setResponse(response);
 
-    if(!responseBody) {
-      throw new ErrorResponseException(StringUtil.format('Could not find index {0}', this.indexName));
+    if(!response.body) {
+      throw new IndexDoesNotExistException(StringUtil.format('Could not find index {0}', this.indexName));
     }
 
-    if (![StatusCodes.Ok, StatusCodes.Accepted].includes(status)) {
+    if (![StatusCodes.Ok, StatusCodes.Accepted].includes(response.statusCode)) {
```
gregolsky, line 38: I noticed you use vscode. I get the following error there. Is there something wrong with my setup or typings?

file: 'file:///c%3A/work/ravendb-nodejs-client/src/Database/Commands/PatchByIndexCommand.ts'
severity: 'Error'
message: 'Property 'includes' does not exist on type 'StatusCode[]'.'
at: '38,49'
source: 'ts'

tryouts/py-json.py
```diff
@@ -0,0 +1,12 @@
+import json
```
gregolsky, line 1: This needs to be removed as well before we release. Please remove tryouts and add it to gitignore.

src/Http/Request/RequestsExecutor.ts
mobilunity-user: i can change to .find()

src/Database/Commands/GetStatisticsCommand.ts
mobilunity-user: yes it's possible in future:
- add those props to basic RavenException
- in auto-thrown exceptions those props will be auto-substituted from json response
- instead of manulally throw new ... we will use ExceptionThrower.throwWith(message, type, response)

src/Database/Commands/PatchByIndexCommand.ts
mobilunity-user: yes, i the same got it. it's a bug of vscode i think - as compiler produce no errors.
i think vscode just doesn't read all typings
anyway .includes() is found in arrays and strings: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/includes?v=control

tryouts/py-json.py
mobilunity-user: ok

src/Database/Commands/PatchByIndexCommand.ts
gregolsky: Yup I know it's there, just wondering why it's being highlighted.

gregolsky: OK I know what's going on. When you target es2016 it's fine. That's because `.includes()` is a part of ES7 spec, not ES6. Not sure if Node 6 is ES2016 though?

gregolsky: @a-serdukov-mobilunity http://node.green/#ES2016

Looking at Node 6.x the only thing we should avoid using is exponentiation (**) operator. Otherwise I think it's fine to change the target to `es2016`.

mobilunity-user: as ts code is compiled to js - it's doensn't matters what Node6 supports.
We just will change target to ES5 and everything will be fine - .includes will be compiled to something like `!!~<array of string>.indexOf(<search for>)` ;)

gregolsky: Well it does matter if we target es2016 - then the features of es2016 are going to be left as is and not replaced with lower ES alternatives - includes stays as includes, right?

Since `includes()` is supported in Node 6 let's just change it to that then and vscode will stop complaining.

PR #14, opened 2017-06-21
build errors fixes

mobilunity-user:


package.json
```diff
@@ -45,10 +45,10 @@
     "gulp-sort": "^2.0.0",
     "gulp-transform": "^1.1.0",
     "gulp-typescript": "^3.1.5",
-    "gulp-uglify-harmony": "https://github.com/jiborobot/gulp-uglify-harmony.git#master",
+    "gulp-uglify-harmony": "git+https://github.com/jiborobot/gulp-uglify-harmony.git#master",
     "mocha": "^3.2.0",
     "rimraf": "^2.6.1",
-    "typescript": "^2.2.1",
+    "typescript": "^2.3.4",
```
gregolsky: Just `2.3.4` - without the dash `^`.

gregolsky: @a-serdukov-mobilunity We have to set "typescript": "2.3.x". Otherwise it's still going to install 2.4 for node 8.

gregolsky (changes requested):

package.json
mobilunity-user: done

package.json
```diff
@@ -45,10 +45,10 @@
     "gulp-sort": "^2.0.0",
     "gulp-transform": "^1.1.0",
     "gulp-typescript": "^3.1.5",
-    "gulp-uglify-harmony": "https://github.com/jiborobot/gulp-uglify-harmony.git#master",
+    "gulp-uglify-harmony": "git+https://github.com/jiborobot/gulp-uglify-harmony.git#master",
     "mocha": "^3.2.0",
     "rimraf": "^2.6.1",
-    "typescript": "^2.2.1",
+    "typescript": "2.3.*",
```
gregolsky: @a-serdukov-mobilunity  Almost :) `*` means *any* version. Make it `~2.3.4`, which means only *patch* number (last one) can increment.

gregolsky (changes requested):

package.json
mobilunity-user: ok

PR #22, opened 2017-07-04
RavenDB-6978: 5th part of changes

mobilunity-user:


src/Documents/Conventions/DocumentConventions.ts
```diff
@@ -47,7 +47,6 @@ export class DocumentConventions {
   readonly defaultUseOptimisticConcurrency:boolean = false;
   readonly maxLengthOfQueryUsingGetUrl = 1024 + 512;
   readonly identityPartsSeparator = "/";
-  private _disableTopologyUpdates: boolean = false;
```
ppekrol, line 50: We need this option in conventions.

src/Http/Request/RequestExecutor.ts
```diff
@@ -12,147 +12,76 @@ import {IRavenObject} from "../../Database/IRavenObject";
...
     };
 
-    setTimeout(() => this.updateFailingNodesStatuses(), 60 * 1000);
-    conventions.topologyUpdatesEnabled && setTimeout(() => this.getReplicationTopology(), 1 * 1000);
+    this._apiKey = apiKey;
+    this._initialUrls = urls;
+    this._initialDatabase = database;
+    this.conventions = conventions;
+    this._lock = Lock.getInstance();
+    this._disableTopologyUpdates = urls.length < 2;
```
ppekrol, line 58: This will not work. Needs to be controlled by the convention only. Imagine a situation when someone will specify 1 url and later on will add a server to the cluster. We want the client to detect that and work properly.

mobilunity-user: @ppekrol Please ,describe how it's handled in C# client

gregolsky: @a-serdukov-mobilunity We need that exposed in `Conventions`. See how we use that to create request executor in c# https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Server/Operations/ServerOperationExecutor.cs#L19

mobilunity-user: it's already refactored after merge witl latest Python version. _disableTopologyUpdates will be re-added again after i'll do ServerOperations

PR #24, opened 2017-07-19
RavenDB-6978

mobilunity-user:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ravendb",
-  "version": "4.0.0",
+  "version": "4.0.0-beta",
```
gregolsky, line 3: I think we're gonna put here `4.0.0-beta-1`

PR #25, opened 2017-07-19
RavenDB-7823

mobilunity-user:


test/session_tests/DocumentStoreTest.ts
```diff
@@ -182,6 +182,16 @@ describe('Document store test', () => {
...
     });
+
+    it('shouldn\'t set id on plain object without id attribute defined', async () => {
+      let product: Product = <Product>{name: "New Product"};    
+
+      session = store.openSession({requestExecutor});      
+      await session.store<Product>(product);
+      await session.saveChanges();
+
+      expect(product.id).to.be.empty;
```
ayende: Why is it expected to be empty here?

test/session_tests/DocumentStoreTest.ts
```diff
@@ -153,13 +153,13 @@ describe('Document store test', () => {
...
+      let product: Product = <Product>{id: null, name: "New Product", uid: null, order: null};    
 
       store.conventions.addDocumentInfoResolver({ resolveDocumentType });
       session = store.openSession({requestExecutor});
       
-      product = await session.store<Product>(product);
+      await session.store<Product>(product);
       await session.saveChanges();
 
       expect(product.id).to.match(/^Product\/\d+(\-\w)?$/);
```
ayende: This should be moved _before_ the `saveChanges`

src/Documents/Conventions/DocumentConventions.ts
```diff
@@ -179,20 +185,23 @@ export class DocumentConventions {
...
+  public setIdOnDocument<T extends Object = IRavenObject>(document: T, id: string, options: IGetSetIDOnDocumentOptions<T> = {}): T {
+    let docType: DocumentType<T> = options.documentType || this.getTypeFromDocument(document);
     const idProperty = this.getIdPropertyName(docType, document);
 
     if ('object' !== (typeof document)) {
       throw new InvalidOperationException("Invalid entity provided. It should implement object interface");
     }
 
-    document[idProperty] = id;
+    if ((false !== options.setIdIfPropertyIsNotDefined) || (idProperty in document)) {
```
ayende: Why do we have this check? `idProperty in document` ?

test/session_tests/DocumentStoreTest.ts
gregolsky: In C# if class does not have an id property defined we don't have anything to populate. We thought it would be good to have similar behavior in JS.

ayende: Probably not a good idea, user can ignore it if they don't need it, and having the option to not have it is fine.
But just putting it by default looks better, I think.

gregolsky: Yes that's the other side of it. JS properties are more flexible on this matter than C#, maybe it's worth it to treat them in a different way. All right, let's add the property if it's not there @a-serdukov-mobilunity

ppekrol: @ayende C# client also removes identity property from JSON when SaveChanges is being called. What behavior should be expected in NodeJS and any other client in regards to that?

ayende: We'll remove that, I think.

ppekrol: @a-serdukov-mobilunity @gregolsky I've talked with @ayende and if possible we would like to have the same behavior in all clients regarding sending the 'Identity property'. So when entity is saved (saveChanges) we want to remove the 'Identity property' from the JSON send, when it is retrieved via session then we want to fill the 'Identity property' in the entity from @metadata[@id]. Is it doable?

gregolsky: @ppekrol I can't see why not. @a-serdukov-mobilunity ?

src/Documents/Conventions/DocumentConventions.ts
mobilunity-user: if javascript object literal  does not have an id property defined we don't have anything to populate.

gregolsky: There's a change of mind on this, we want to create that property in object literal if it's not defined and reverse the option `setIdIfPropertyIsNotDefined` to `setIdOnlyIfPropertyIsDefined` or sth like that. Please see other comments.

gregolsky:
Can we add `setIdOnlyIfPropertyIsDefined` option to conventions (false by default)? Just to give the user a choice.

PR #28, opened 2017-08-21
add rql and patch

mobilunity-user:


gregolsky:
@mobilunity-user @a-serdukov-mobilunity Could you guys comment in English please?

mobilunity-user:
Yes,sorry, this were the comments for me

src/Database/Commands/DeleteByQueryCommand.ts
```diff
@@ -1,30 +1,36 @@
...
+
+    this.payload = this.query.toJson();
+
   }
 
   public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | void {
-    const result: IRavenResponse = <IRavenResponse>super.setResponse(response);    
+    const result: IRavenResponse = <IRavenResponse>super.setResponse(response);
 
     if (!response.body) {
```
gregolsky: Is the empty body the only indication that there's no such index? I'd check the status code as well.

src/Database/Commands/PatchByQueryCommand.ts
```diff
@@ -23,22 +23,24 @@ export class PatchByIndexCommand extends IndexQueryBasedCommand {
...
 
   public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | void {
     const result: IRavenResponse = <IRavenResponse>super.setResponse(response);
 
-    if(!response.body) {
-      throw new IndexDoesNotExistException(StringUtil.format('Could not find index {0}', this.indexName));
-    }
-
     if (![StatusCodes.Ok, StatusCodes.Accepted].includes(response.statusCode)) {
       throw new ErrorResponseException('Invalid response from server');
```
gregolsky: Please include details what's wrong with that response - sth like (*Invalid response from server: statusCode message\n body*) probably. It would probably be worth it to include those in ErrorResponseException and format message using them.

src/Documents/RQL/RQLOperator.ts
```diff
@@ -0,0 +1,13 @@
+export type rqlOperator = 'greaterThan' | 'lessThan' | 'search' | 'equals' | 'between' | 'endsWith' | 'startsWith' | 'in' | 'boost';
+
+export class rqlOperators {
+  public static readonly greaterThan: rqlOperator = 'greaterThan';
```
gregolsky: Since these are constants we probably should use SNAKE_CASE for these. Also *rqlOperators* - class names should be PascalCased.

test/raven_command_tests/ByIndexCommandsTest.ts
```diff
@@ -45,56 +46,63 @@ describe('IndexBasedCommand tests', () => {
     });
 
     patch = new PatchRequest("Name = 'Patched';");
-    
+    query = `from index 'Testing_Sort' where exists(Name)`;
```
gregolsky: It would be nice to have tests made on data that actually mean something like Orders, Products etc. You can use docs from `Northwind`, which data model is fairly popular. *Testings*, *Testing_Sort* or *Universal* does not mean anything. Remember that tests are one of primary sources of knowledge about this project. Show what's possible and what's not. It's important they're understandable, logical and easy to read.

Link to Northwind http://4.live-test.ravendb.net/studio/index.html#databases/documents?&database=Northwind

test/raven_command_tests/ByIndexCommandsTest.ts
```diff
@@ -45,56 +46,63 @@ describe('IndexBasedCommand tests', () => {
...
-          expect(response).not.to.be.null;
-          expect((response as IRavenResponse).Result.Total).not.to.be.lessThan(50);
-        })
-      );
+          .send(PatchByQueryOperations)
+          .then((response: IRavenResponse) => {
+            expect(response).not.to.be.null;
+            expect((response as IRavenResponse).Result.Total).not.to.be.lessThan(1);
+          }, err => {
+            console.log(err);
```
gregolsky: Promise should be rejected, not just log error. This kind of code hides failing tests.

test/session_tests/DocumentLoadTest.ts
```diff
@@ -102,13 +102,13 @@ describe('Document load test', () => {
       expect(company.product).to.be.an.instanceOf(Product);
     });
 
-    it('should load with includes', async () => {
-      session = store.openSession();
-      
-      await session.load<Order>("Order/105", Order, ["product_id"]);
-      await session.load<Product>("Product/101");
+    // it('should load with includes', async () => {
```
gregolsky: Don't we support includes anymore?

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,289 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+        nestedObjectTypes: {product: Product}
+      }).whereEquals<string>('name', 'withNesting').get();
 
-    it('should query by whereIn', async() => {
-      const results: Product[] = await store.openSession().query<Product>({
-        documentType: Product
-      }).whereIn<string>('name', ['test101', 'test107', 'test106']).get();
-      
-      expect(results).to.have.lengthOf(4);      
+      expect(results[0].product).to.be.instanceOf(Product);
```
gregolsky: Assertions in these tests are useless. We just check we got something, but don't care what. Please add checks for the number of results you expect, the content of objects etc. This comment applies to every `it()` in this file.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,289 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+      });
+
+  it('should query with whereIn', async () => {
+    const results: Universal[] = await store.openSession()
+      .query<Universal>({
+        indexName: 'Universals',
+        WaitForNonStaleResults: true
+      })
+      .whereIn('name', 'withNesting').get();
+    expect(results[0]).to.be.a('object');
```
gregolsky: What if results is an array like`[{}]`.. We don't check anything in here.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,289 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
-      expect(session.numberOfRequestsInSession).to.equals(1);            
-    });
+  it('should query with orderByDESC', async () => {
+    const results: Universal[] = await store.openSession()
+      .query<Universal>({
+        indexName: 'Universals',
+        WaitForNonStaleResults: true
+      })
+      .orderBy<string>('name', 'DESC').get();
+    expect(results[0]).to.be.a('object');
```
gregolsky: It's orderBy test and you don't even check if the results you got are ordered in any way.

gregolsky (changes requested):
Please check out my comments and apply the changes. All tests are passing.

@arekpalinski @ppekrol Could one of you take a look from the RQL impl validity point of view please?

src/Documents/RQL/QueryBuilder.ts
```diff
@@ -0,0 +1,164 @@
...
+  public select(fields) {
+
+    if (TypeUtil.isArray(fields)) {
+      this._select = fields.join(', ');
+    }
+    else if (fields) {
+      this._select = fields;
+    }
+    else {
+      this._select = '*';
```
ayende: If there is no fields, we can emit the select entirely.

src/Database/Indexes/IndexQuery.ts
```diff
@@ -11,19 +13,16 @@ export class IndexQuery {
   private _defaultOperator?: QueryOperator = null;
   private _waitForNonStaleResults: boolean = false;
   private _waitForNonStaleResultsTimeout?: number = null;
+  private _buffer:string = null;
 
-  constructor(query: string = '', pageSize: number = 128, skippedResults: number = 0,
-    defaultOperator?: QueryOperator, options: IOptionsSet = {}
-  ) {
+  constructor(query: string = '', pageSize: number = 128, skippedResults: number = 0, options: IOptionsSet = {}) {
```
ppekrol: PageSize by default now should be int.MaxValue (https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Queries/IndexQuery.cs#L183)

 but we should only send it if it was changed by the user:
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Queries/IndexQuery.cs#L211

src/Documents/RQL/QueryBuilder.ts
```diff
@@ -0,0 +1,164 @@
...
+      case rqlOperators.equals:
+        (conditionValue === null || conditionValue === 'null') ?
+          this.rqlText = StringUtil.format(`{0}={1} `, conditionField, conditionValue) :
+          this.rqlText = StringUtil.format(`{0}='{1}' `, conditionField, conditionValue);
+        break;
+      case rqlOperators.between:
+        this.rqlText = StringUtil.format(`{0} BETWEEN {1} AND {2}`, conditionField, conditionValue.start, conditionValue.end);
+        break;
+      case rqlOperators.startsWith:
+        this.rqlText = StringUtil.format(`StartsWith({0}, '{1}')`, conditionField, conditionValue);
```
ppekrol: can be lowercased: `startsWith`

src/Documents/RQL/QueryBuilder.ts
```diff
@@ -0,0 +1,164 @@
...
+        break;
+      case rqlOperators.endsWith:
+        this.rqlText = StringUtil.format(`EndsWith({0}, '{1}')`, conditionField, conditionValue);
```
ppekrol: can be lowercased: `endsWith`

src/Documents/RQL/QueryBuilder.ts
```diff
@@ -0,0 +1,164 @@
...
+        break;
+      case rqlOperators.in:
+        this.rqlText = StringUtil.format(`{0} IN ('{1}')`, conditionField, conditionValue);
+        break;
+      case rqlOperators.search:
+        this.rqlText = StringUtil.format(`search({0},'{1}') `, conditionField, conditionValue);
+        break;
+      case rqlOperators.boost:
+        this.rqlText = StringUtil.format(`boost({0} = '{1}', {2})`, conditionField.boostField, conditionField.boostValue, conditionValue);
```
ppekrol: `boost` needs to support any expression `boost(expression, boostValue)`
e.g. `boost(Age > 20, 5)`

src/Documents/RQL/QueryBuilder.ts
```diff
@@ -0,0 +1,164 @@
...
+        break;
```
ppekrol: add support for `exact(expression)`

ppekrol: add support for query parameters.
e.g. `FROM Users Where Age > :p0`

JSON would look like:
```
{
	"Query": "FROM Users Where Age > :p0",
	"QueryParameters": {
		"p0" : 20
	}
}
```

src/Documents/Session/DocumentQuery.ts
```diff
@@ -45,299 +48,330 @@ export class DocumentQuery<T> extends Observable implements IDocumentQuery<T> {
...
+
   }
 
-  protected addLuceneCondition<C extends LuceneConditionValue>(fieldName: string, condition: C,
-    operator?: LuceneOperator, escapeQueryOptions: EscapeQueryOption = EscapeQueryOptions.EscapeAll,
-    boost: number = 1
-  ): DocumentQuery<T> {
-    const luceneCondition: string = LuceneBuilder.buildCondition<C>(this.session.conventions, fieldName,
-      condition, operator, escapeQueryOptions, boost);
+  public whereBetweenOrEqual<V extends RQLValue>(field, start, end, orName, orValue) {
```
ppekrol: we dropped support for this

src/Documents/RQL/QueryBuilder.ts
ppekrol: add support for `OpenSubclause` and `CloseSubclause`

ppekrol: check what methods we are supporting here https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/IDocumentQueryBase.cs

src/Documents/RQL/QueryBuilder.ts
```diff
@@ -0,0 +1,164 @@
...
+    return this;
+
+  }
+
+
+  public whereRaw(rqlCondition) {
+
+    let addNot: string = this.negate ? 'NOT ' : '';
+
+    this._where += StringUtil.format(`{0} {1} {2}`, this.operator, addNot, rqlCondition);
```
ppekrol: concatenating strings might not be a good idea, because you need to be aware what statements are before the one that you are trying to insert (in some cases).

Check: https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs#L1695-L1713
or 
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs#L1640-L1669

ppekrol: Consider using list of tokens
https://github.com/ravendb/ravendb/tree/v4.0/src/Raven.Client/Documents/Session/Tokens
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs#L113
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/AbstractDocumentQuery.cs#L1322-L1327

ppekrol (changes requested):

src/Documents/RQL/QueryBuilder.ts
mobilunity-user: you're right, it's a good idea and we're think already about this. But it may take more time so we decided to do RQL builder closer to current implementation (using concatenation) for released it ASAP (as we got also a lot of more changes / updates need to fix)
We can plan to imlement token-based builder in next PR (we'll add task  in youtrack)

test/raven_command_tests/ByIndexCommandsTest.ts
mobilunity-user: Yes we can do it after we'll implement the most important features. As you remember we already discussed that using data fixtures of real databases is an better option then put documents manually

src/Database/Commands/PatchByQueryCommand.ts
mobilunity-user: We have special code responses from the server, which we can include in the error or I just enter the error code in the message?

gregolsky: Let's use `ExceptionThrower` class we have for this kind of situations.

src/Database/Commands/DeleteByQueryCommand.ts
```diff
@@ -26,7 +27,7 @@ export class DeleteByQueryCommand extends QueryBasedCommand  implements IJSonSer
   public setResponse(response: IResponse): IRavenResponse | IRavenResponse[] | void {
     const result: IRavenResponse = <IRavenResponse>super.setResponse(response);
 
-    if (!response.body) {
+    if (!response.body && StatusCodes.isError(response.statusCode)) {
       throw new IndexDoesNotExistException('Could not find index');
```
gregolsky: Index does not exist only if you get 404 Not found. Other status codes, even if 4xx or 5xx may not mean that. Also please remove the `!response.body` check from the condition here. Let's do the following
1. if 404 => index does not exist
2. StatusCodes.isError => throw more generic exception (ErrorResponseException for example)

src/Documents/Session/IDocumentQuery.ts
```diff
@@ -1,50 +1,48 @@
...
-  andAlso(): IDocumentQuery<T>;
-  orElse(): IDocumentQuery<T>;
-  negateNext(): IDocumentQuery<T>;
-  take(docsCount: number): IDocumentQuery<T>;
-  skip(skipCount: number): IDocumentQuery<T>;
+  search(from: string, searchTerms: string | string[], boostField, boostValue, boostExpression, count): IDocumentQuery<T>;
+  where(conditions: IDocumentQueryConditions);
+  whereEquals<V extends RQLValue>(field: string, value: V): IDocumentQuery<T>;
+  endsWith<V extends RQLValue>(field: string, value: V): IDocumentQuery<T>;
+  exact<V extends RQLValue>(field: string, value: V): IDocumentQuery<T>;
```
gregolsky, line 34: Exact should not be a method. It's just an option of whereEquals. Please see https://github.com/ravendb/ravendb/blob/v4.0/test/SlowTests/MailingList/Troy.cs#L50

src/Documents/Session/IDocumentQuery.ts
```diff
@@ -1,50 +1,48 @@
...
+  startsWith<V extends RQLValue>(field: string, value: V): IDocumentQuery<T>;
+  whereIn<V extends RQLValue>(field: string, value: V): IDocumentQuery<T>;
+  whereBetween<V extends RQLValue>(field: string, start?: V, end?: V, orName?, orValue?): IDocumentQuery<T>;
+  whereGreaterThan<V extends RQLValue>(field: string, value: V): IDocumentQuery<T>;
+  whereGreaterThanOrEqual<V extends RQLValue>(field: string, value: V, orName?, orValue?): IDocumentQuery<T>;
+  whereEqualsAndOr<V extends RQLValue>(fieldFirst: string, valueFirst: string, fieldSecond: string, valueSecond: number, fieldThird: string, valueThird: number): IDocumentQuery<T>;
```
gregolsky, line 40: There's no such query method in C# client. We'd like Client API to be consistent across platforms. Only query methods listed in https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/IDocumentQueryBase.cs
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/IDocumentQuery.cs

should be available.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
-      expect(results).to.have.lengthOf(4);      
+    it('should query with orderBy as string', async () => {
+      const results: Universal[] = await store.openSession()
+        .query<Universal>({
+          indexName: 'Universals',
+          WaitForNonStaleResults: true
+        })
+        .orderBy<string>('order', 'ASC').get();
+
+      expect(results[1].order).to.be.equal(10);
```
gregolsky, line 105: When testing `orderBy` method you should iterate over the list of results and check whether they are ordered.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+      const results: Universal[] = await store.openSession()
+        .query<Universal>({
+          indexName: 'Universals',
+          WaitForNonStaleResults: true
+        })
+        .whereGreaterThanOrEqual<number>('order',6, 'name', 'withNesting').get();
+
+      expect(results).to.have.lengthOf(10);
+
+        results.map(function(obj) {
```
gregolsky, line 136: Assertions are wrong. You test `whereGreaterThanOrEqual` but check whether results contain `name` property set to 'withNesting'. You should verify that every result has property 'order' greater or equal to 6.

Also what are those additional parameters passed to `whereGreaterThanOrEqual` (`...'name', 'withNesting')`?

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+          indexName: 'Universals',
+          WaitForNonStaleResults: true
+        })
+        .whereLessThanOrEqual<number>('order',6, 'name', 'withNesting').get();
+
+      expect(results).to.have.lengthOf(10);
+
+      results.map(function(obj) {
+
+        expect(obj).to.include({'name': 'withNesting'});
```
gregolsky, line 156: Wrong assertions and redundant parameters for `whereLessThanOrEqual` - see comment above.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+
+    });
+
+    it('should query with whereArray', async () => expect(
+      store.openSession().query<Universal>({
+        indexName: 'Universals'
+      }).where({
+        name: [4, 6, 90]
+      }).get()
+      ).to.be.fulfilled
```
gregolsky, line 194: It's just checking whether the query succeeded. Its results should be verified with proper assertions.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
-    it('should query with index', async() => {
-      const results: Product[] = await store.openSession().query<Product>({
-        documentType: Product,
-        indexName: 'Testing_Sort',
-        waitForNonStaleResults: true
-      }).where({
-        uid: [4, 6, 90]
-      }).get();
+    it('should fail query with non-existing index', async () => {
+      const results = await store.openSession().query<Universal>({
```
gregolsky, line 205: Since we throw exception when index does not exist, where's the verification part of that fact in this test?

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+        indexName: 'none'
+      }).whereIn('Tag', 'Products').get();
+
+      expect(results).to.have.lengthOf(0);
+
+    });
+
+    it('should query by @all_docs index', async () => {
```
gregolsky, line 213: Either test description or its body is wrong. I can't see reference to `@all_docs` anywhere here.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
-      }).whereBetweenOrEqual<number>('uid', 2, 4).get();
-      
-      expect(results).to.have.lengthOf(3);
+    it('should query with andAlso', async () => {
+      const results = await store.openSession()
+        .query<Universal>({
+          indexName: 'Universals',
+          WaitForNonStaleResults: true
+        })
+        .andAlso();
```
gregolsky, line 267: I believe you're missing something past the `andAlso()` call. Left like this it does not affect the query at all.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
-      }).whereNotNull('order').get();
-       
-      expect(_.some(results, (result: IRavenObject) => TypeUtil.isNone(result.order))).to.be.false; 
+    it('should query with orElse', async () => {
+      const results = await store.openSession()
+        .query<Universal>({
+          indexName: 'Universals',
+          WaitForNonStaleResults: true
+        })
+        .orElse();
```
gregolsky, line 280: same as above

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
+        .whereLessThan<string>('order', ':p0').get();
+
+      expect(results).to.have.lengthOf(5);
 
-      await session.query({
-        waitForNonStaleResults: true, 
-        includes: ['product_id']
-      }).where({uid: 92}).get();
+      results.map(function (obj) {
+        expect(obj.order).to.not.be.greaterThan(6);
```
gregolsky, line 364: That's what I was talking about when commenting `whereGreaterOrEquals` 👍 Please add this kind of assertions where needed.

test/session_tests/DocumentQueryTest.ts
```diff
@@ -8,220 +8,493 @@ import {RequestExecutor} from "../../src/Http/Request/RequestExecutor";
...
     });
-  });  */
+
+    it('should query by search', async () => {
+      const results: Universal[] = await store.openSession()
+        .query<Universal>({
+          indexName: 'Universals',
+          WaitForNonStaleResults: true
+        })
+        .search('name', 'withNesting', 'order', 2, '=', 2).get();
```
gregolsky, line 410: This call is hardly understandable. What's happening here?

gregolsky: I believe it's made this way due to lack of tokenization and builder.. This however makes API inconsistent. We need it like this:
https://github.com/ravendb/ravendb/blob/v4.0/src/Raven.Client/Documents/Session/IDocumentQueryBase.cs#L349
```
 public search<V extends RQLValue>(field, value, boostField, boostValue, count) {

    this._builder
      .from(this.indexName, this.fromCollection)
      .where('search', field, value)
      .or
      .where('boost', {boostField: boostField, boostValue: boostValue}, count);

    return this;
  }
```

mobilunity-user:
closed until i'll complete code review and put latest part of requested changes
they'll be in next PR

test/session_tests/DocumentQueryTest.ts
mobilunity-user: If the index does not exist, then we`re not getting an error, but just an empty array

mobilunity-user: If the index does not exist, then we are not getting an error, but just an empty array

ayende: This isn't a query on an index that doesn't exists. It is a dynamic query that gives no results.
You need to specify the index name to get it there.

PR #40, opened 2017-10-17
RDBC-29 initial Hilo value in examples from README set to 0 (how it should be)

mobilunity-user:


gregolsky:
I think it's a bug we start with 0 in Node.js client. C# and the Studio store documents and start with 1.

ayende:
Yes, we should start with 1

PR #41, opened 2017-10-17
RDBC 28

mobilunity-user:


gregolsky (changes requested):
We need to revert "initial Hilo value set to 0" - c72642a.
See: http://issues.hibernatingrhinos.com/issue/RDBC-35

README.md
```diff
@@ -32,7 +32,7 @@ const session = store.openSession();
 4. Call `saveChanges()` when you'll finish working with a session
 ```javascript
 session
- .load('Users/1-A')
+ .load('Users/0-A')
```
ayende: This is wrong. We should never generate 0 as id.

ayende (changes requested):

PR #44, opened 2017-10-26
Rdbc 45

mobilunity-user:
RDBC-45 Node.js client not setting metadata properly.


ayende:
Please also create a test for this

PR #56, opened 2017-11-30
fix RDBC-68 HiLo is not incremented properly

mobilunity-user:


gregolsky:
can we add a test for this please, so it doesn't break in the future?

gregolsky:
Sample app works fine now but, HiLo tests are failing.
```
1) HiLo generator test HiLo should requests new range:

      AssertionError: expected 1 to be above 32
      + expected - actual

      -1
      +32

      at Object.<anonymous> (.build\test\session_tests\HiloGeneratorTest.js:59:69)
      at next (native)
      at fulfilled (.build\test\session_tests\HiloGeneratorTest.js:7:58)



npm ERR! Test failed.  See above for more details.
```

PR #64, opened 2018-03-03
Fix spatial order tokens

configurator:


ayende:
@configurator Thanks for that, haven't heard from you in a while!

configurator:
:wave:

PR #83, opened 2018-05-25
advanced patching, replication and tests

gregolsky:


gregolsky:
License variable is not injected due to security reasons, hence the failure. I'll look into it later today.

PR #101, opened 2018-08-09
RDBC-213 Support storing metadata using @metadata field on entities

gregolsky:


src/Documents/Session/DocumentInfo.ts
```diff
@@ -36,13 +36,17 @@ export class DocumentInfo {
...
         }
 
         const newDocumentInfo = new DocumentInfo();
         newDocumentInfo.id = id;
-        newDocumentInfo.document = document;
-        newDocumentInfo.metadata = metadata;
+      
+        newDocumentInfo.document = {};
+        newDocumentInfo.metadata = {};
+        Object.assign(newDocumentInfo.document, document);
```
gregolsky, line 47: @Danielle9897 Java does not clone here - just assigns. Why do we need make a new obj?
https://github.com/ravendb/ravendb-jvm-client/blob/v4.0/src/main/java/net/ravendb/client/documents/session/DocumentInfo.java#L164

src/Documents/Session/EntityToJson.ts
```diff
@@ -71,19 +71,32 @@ export class EntityToJson {
...
+        documentInfo.metadata[CONSTANTS.Documents.Metadata.NESTED_OBJECT_TYPES] = typeInfo.nestedTypes;
+        documentInfo.metadata[CONSTANTS.Documents.Metadata.RAVEN_JS_TYPE] = typeInfo.typeName;
 
-        if (documentInfo.metadata 
-            && Object.keys(documentInfo.metadata).length > 0) {
+        if (documentInfo.metadataInstance) {
             setMetadata = true;
-            Object.assign(metadataNode, documentInfo.metadata);
-        }
+            metadataNode = JSON.parse(JSON.stringify(documentInfo.metadataInstance));
```
gregolsky: @Danielle9897  Please add `ObjectUtil` class in Utility directory with a `.deepClone(obj)` method and replace all occurences of `JSON.parse(JSON.stringify(x))`.

src/Documents/Session/EntityToJson.ts
```diff
@@ -71,19 +71,32 @@ export class EntityToJson {
...
         let setMetadata: boolean = false;
-        const metadataNode: object = {};
+        let metadataNode: object = {};
+        
+        documentInfo.metadata[CONSTANTS.Documents.Metadata.NESTED_OBJECT_TYPES] = typeInfo.nestedTypes;
+        documentInfo.metadata[CONSTANTS.Documents.Metadata.RAVEN_JS_TYPE] = typeInfo.typeName;
 
-        if (documentInfo.metadata 
-            && Object.keys(documentInfo.metadata).length > 0) {
+        if (documentInfo.metadataInstance) {
```
gregolsky, line 81: @Danielle9897  Can we switch `if` and `else if` blocks so that it's more similar to JVM codebase?

https://github.com/ravendb/ravendb-jvm-client/blob/v4.0/src/main/java/net/ravendb/client/documents/session/EntityToJson.java#L77

First check for `documentInfo.metadata && Object.keys(documentInfo.metadata).length > 0` and then `documentInfo.metadataInstance`.

src/Documents/Session/DocumentInfo.ts
Danielle9897: The test fails without this!

src/Documents/Session/EntityToJson.ts
Danielle9897: ok

gregolsky:
Closing this one for now until we figure out the discrepancies between client implementations.

ayende:
I don't follow what the issue here is?

gregolsky:
There are some differences between Java and node.js clients in those lines and it's going to be fixed when doing RDBC-147 - sync up to Java client 4.0.2.

PR #102, opened 2018-08-12
RDBC-213 Support storing metadata using @metadata field on entities. …

Danielle9897:
…NOTE: this is a reduced PR, w/o the new test that checks updating the metadata


gregolsky:
Please add the new test with metadata updates, but make it skipped for the time being.

gregolsky:
I rerun PR checks, it failed due to `npm install` failure.

PR #115, opened 2018-09-11
changes API

ml054:
sync with java: 51a50e3a6c58deb06778f5d5da50a4cb5d03e583, e4e6c42fafa58f5a3ecaf69b76e9300a92639f4c, 912c7ffd0939d64a5fb71660f48e63639f63ec94

### DIFF
--- a/src/Documents/Changes/ChangesObservable.ts
+++ b/src/Documents/Changes/ChangesObservable.ts
@@ -1,0 +1,96 @@
+import {ChangesType, IChangesConnectionState} from "./IChangesConnectionState";
+import {IChangesObservable} from "./IChangesObservable";
+
+export class ChangesObservable<T, TConnectionState extends IChangesConnectionState<any>>
+    implements IChangesObservable<T> {
+
+    private readonly _type: ChangesType;
+    private readonly _connectionState: TConnectionState;
+    private readonly _filter: (val: T) => boolean;
+    private readonly _subscribers: Set<(value: T) => void> = new Set();
+    private readonly _errorSubscribers: Set<(error: Error) => void> = new Set();
+
+    private _sendHandler: (val: T) => void;
+    private _errorHandler: (val: Error) => void;
+
+    public constructor(type: ChangesType, connectionState: TConnectionState, filter: (val: T) => boolean) {
+        this._type = type;
+        this._connectionState = connectionState;
+        this._filter = filter;
+    }
+
+    public on(event: "data", handler: (value: T) => void);
+    public on(event: "error", handler: (error: Error) => void);
+    public on(event: "data" | "error", handler: ((value: T) => void) | ((error: Error) => void)) {
+        switch (event) {
+            case "data":
+                // since allow multiple subscriptions on single object we cant register it multiple times
+                // to avoid duplicates in notification
+                if (!this._sendHandler) {
+                    // register shared handler
+                    this._sendHandler = (payload: T) => this.send(payload);
+                    this._connectionState.addOnChangeNotification(this._type, this._sendHandler);
+                }
+
+                this._subscribers.add(handler as (value: T) => void);
+                this._connectionState.inc();
+                break;
+            case "error":
+                if (!this._errorHandler) {
+                    // register shared error handler
+                    this._errorHandler = (ex: Error) => this.error(ex);
+                    this._connectionState.addOnError(this._errorHandler);
+                }
+
+                this._errorSubscribers.add(handler as (error: Error) => void);
+                break;
+        }
+    }
+
+    public off(event: "data", handler: (value: T) => void);
+    public off(event: "error", handler: (error: Error) => void);
+    public off(event: "data" | "error", handler: ((value: T) => void) | ((error: Error) => void)) {
+
+        this._connectionState.dec();
+
+        switch (event) {
+            case "data":
+                this._subscribers.delete(handler as (value: T) => void);
+                if (!this._subscribers.size) {
+                    // no more subscribers left - remove from parent
+                    this._connectionState.removeOnChangeNotification(this._type, this._sendHandler);
+                    this._sendHandler = undefined;
+                }
+
+                break;
+            case "error":
+                this._errorSubscribers.delete(handler as (error: Error) => void);
+                if (!this._errorSubscribers.size) {
+                    this._connectionState.removeOnError(this._errorHandler);
+                    this._errorHandler = undefined;
+                }
+                break;
+        }
+    }
+
+    public send(msg: T): void {
+        try {
+            if (!this._filter(msg)) {
+                return;
+            }
+        } catch (e) {
+            this.error(e);
+            return;
+        }
+
+        this._subscribers.forEach(x => x(msg));
+    }
+
+    public error(e: Error): void {
+        this._errorSubscribers.forEach(x => x(e));
+    }
+
+    public ensureSubscribedNow(): Promise<void> {
+        return this._connectionState.ensureSubscribedNow();
+    }
+}
--- a/src/Documents/Changes/DatabaseChanges.ts
+++ b/src/Documents/Changes/DatabaseChanges.ts
@@ -20,8 +20,6 @@ import {ObjectUtil} from "../../Utility/ObjectUtil";
 
 export class DatabaseChanges implements IDatabaseChanges {
 
-    private static readonly ON_CONNECTION_STATUS_CHANGED = "connectionStatus";
-
     private _emitter = new EventEmitter();
     private _commandId: number = 0;
     private _onConnectionStatusChangedWrapped: () => void;
@@ -51,7 +49,7 @@ export class DatabaseChanges implements IDatabaseChanges {
         this._tcs = PromiseUtil.defer<IDatabaseChanges>();
         this._onDispose = onDispose;
         this._onConnectionStatusChangedWrapped = () => this._onConnectionStatusChanged();
-        this._emitter.on(DatabaseChanges.ON_CONNECTION_STATUS_CHANGED, this._onConnectionStatusChangedWrapped);
+        this._emitter.on("connectionStatus", this._onConnectionStatusChangedWrapped);
         this._task = this._doWork();
     }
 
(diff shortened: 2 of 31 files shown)


src/Documents/Changes/ChangesObservable.ts
```diff
@@ -0,0 +1,96 @@
...
+    private readonly _connectionState: TConnectionState;
+    private readonly _filter: (val: T) => boolean;
+    private readonly _subscribers: Set<(value: T) => void> = new Set();
+    private readonly _errorSubscribers: Set<(error: Error) => void> = new Set();
+
+    private _sendHandler: (val: T) => void;
+    private _errorHandler: (val: Error) => void;
+
+    public constructor(type: ChangesType, connectionState: TConnectionState, filter: (val: T) => boolean) {
+        this._type = type;
```
gregolsky, line 17: do we need to validate args?

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+            options = certificate.toWebSocketOptions();
+        }
+
+        return new WebSocket(url, options);
+    }
+
+    private _onConnectionStatusChanged() {
+        const acquiredSemContext = acquireSemaphore(this._semaphore);
+
+        BluebirdPromise.resolve(acquiredSemContext.promise)
```
gregolsky, line 71: I think this is going to warn about *unreturned* promises. Also can the `then()` content blow up? We should account for that.

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+    public dispose(): void {
+        Array.from(this._confirmations.values()).forEach(confirmation => confirmation.reject());
+
+        this._isCancelled = true;
+        this._client.close();
+
+        this._counters.clear();
+
+        this._emitter.emit(DatabaseChanges.ON_CONNECTION_STATUS_CHANGED);
+        this._emitter.removeListener(DatabaseChanges.ON_CONNECTION_STATUS_CHANGED,
```
gregolsky: I think we should remove all event listeners (`removeAllListeners()`) from our emitter here and null it.

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+    public async ensureConnectedNow(): Promise<void> {
+        await this._tcs;
+    }
+
+    public forIndex(indexName: string): IChangesObservable<IndexChange> {
+        const counter = this._getOrAddConnectionState("indexes/" + indexName,
+            "watch-index", "unwatch-index", indexName);
+
+        return new ChangesObservable<IndexChange, DatabaseConnectionState>("Index", counter,
+            notification => notification.name && notification.name.toLocaleLowerCase()
```
gregolsky: can you move stuff from && to new line

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+    public forDocumentsStartingWith(docIdPrefix: string): IChangesObservable<DocumentChange> {
+        const counter = this._getOrAddConnectionState("prefixes/" + docIdPrefix,
+            "watch-prefix", "unwatch-prefix", docIdPrefix);
+        return new ChangesObservable<DocumentChange, DatabaseConnectionState>("Document", counter,
+            notification => notification.id
+                && notification.id.toLocaleLowerCase().startsWith(docIdPrefix.toLocaleLowerCase()));
+    }
+
+    public forDocumentsInCollection<T extends object>(type: ObjectTypeDescriptor<T>);
+    public forDocumentsInCollection(collectionName: string): IChangesObservable<DocumentChange>;
```
gregolsky: please move the one using collection name up. That's more natural for JS.

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+        const counter = this._getOrAddConnectionState("indexes/" + indexName,
+            "watch-index", "unwatch-index", indexName);
+
+        return new ChangesObservable<IndexChange, DatabaseConnectionState>("Index", counter,
+            notification => notification.name && notification.name.toLocaleLowerCase()
+                === indexName.toLocaleLowerCase());
+    }
+
+    public get lastConnectionStateException(): Error {
+        for (const counter of Array.from(this._counters.values())) {
```
gregolsky: isn't `values()` result an iterable? do we need to Array.from()?

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+
+        if (!collectionName) {
+            throwError("InvalidArgumentException", "CollectionName cannot be null");
+        }
+
+        const counter = this._getOrAddConnectionState("collections/" + collectionName,
+            "watch-collection", "unwatch-collection", collectionName);
+
+        return new ChangesObservable<DocumentChange, DatabaseConnectionState>("Document", counter,
+            notification => notification.collectionName && collectionName.toLocaleLowerCase()
```
gregolsky: pls break line before &&

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+                    setTimeout(() => this._doWorkInternal(url), 1000);
+                }
+
+                Array.from(this._confirmations.values()).forEach(v => v.reject());
+                this._confirmations.clear();
+            });
+
+            this._client.on("message", (data: WebSocket.Data) => {
+                this._processChanges(data as string);
+            });
```
gregolsky, line 326: Do we need to pingpong? https://github.com/websockets/ws#how-to-detect-and-close-broken-connections

src/Documents/Changes/DatabaseChanges.ts
```diff
@@ -0,0 +1,400 @@
...
+                    default:
+                        const value = message.Value;
+                        const transformedValue = ObjectUtil.transformObjectKeys(value,  { defaultTransform: "camel" });
+                        this._notifySubscribers(type, transformedValue, Array.from(this._counters.values()));
+                        break;
+                }
+            }
+        } catch (err) {
+            this._notifyAboutError(err);
+            throwError("ChangeProcessingException", err);
```
gregolsky: I think you missed the message arg.

test/Ported/Server/Documents/Notifications/ChangesTest.ts
```diff
@@ -0,0 +1,308 @@
...
+        await changes.ensureConnectedNow();
+
+        const observable = changes.forDocumentsStartingWith("users/");
+        await observable.ensureSubscribedNow();
+
+        const handler = (change: DocumentChange) => changesList.push(change);
+
+        try {
+            observable.on("data", handler);
+            observable.on("error", e => throwError("InvalidOperationException", e.message));
```
gregolsky, line 214: I think throwing inside of the handler won't bubble up to try block, but will silently fail there. Since it's a test you'd need to do `done(err)` or similar.

test/Ported/Server/Documents/Notifications/ChangesTest.ts
```diff
@@ -0,0 +1,308 @@
...
+
+            documentChange = await changesList.poll(2000);
+            assert.ok(documentChange);
+            assert.strictEqual(documentChange.id, "users/2");
+        } finally {
+            observable.off("data", handler);
+        }
+    });
+
+    it.skip("doesn't crush the server if listening on non existing database", async () => {
```
gregolsky: why skip?

src/Documents/Changes/ChangesObservable.ts
ml054: I don't think so, we call this from internal functions.

src/Documents/Changes/DatabaseChanges.ts
ml054: it doesn't warn about this. There is nothing which can throw in then statement.

src/Documents/Changes/DatabaseChanges.ts
ml054: the effect is the same, but when we keep current approach it will be more bulletproof when c# code change, so I prefer to leave it.

src/Documents/Changes/DatabaseChanges.ts
ml054: done

src/Documents/Changes/DatabaseChanges.ts
ml054: sure, done

src/Documents/Changes/DatabaseChanges.ts
ml054: it is, thanks .

src/Documents/Changes/DatabaseChanges.ts
ml054: done

src/Documents/Changes/DatabaseChanges.ts
ml054: no - server is sending empty messages as heartbeats

src/Documents/Changes/DatabaseChanges.ts
ml054: Copied from java and c#. In general it means handler for changes api throw an exception.

test/Ported/Server/Documents/Notifications/ChangesTest.ts
ml054: skip removed.

src/Documents/Changes/DatabaseChanges.ts
gregolsky: OK, but you cannot omit it when using this function anyway - inner error goes into the third arg and message - 2nd.

test/Ported/Server/Documents/Notifications/ChangesTest.ts
ml054: I've added message.

PR #130, opened 2018-09-24
RDBC-230 deep copy document to entity in document info

gregolsky:


gregolsky:
need to fix it for new patch tests

PR #155, opened 2018-10-09
update stream-json dep; use values only json parsers

gregolsky:


gregolsky:
Checks failed due to Travis/Ubuntu software sources repo issues. Tested manually.
```
  385 passing (3m)
  9 pending
```

PR #156, opened 2018-10-10
RDBC-234 test for lazy loading

ml054:


test/Ported/Lazy/LazyTest.ts
```diff
@@ -151,4 +151,47 @@ describe("LazyTest", function () {
             assert.strictEqual(user.lastName, "Oren");
         }
     });
+
+    it.only("doesn't lazy load already loaded values", async () => {
```
gregolsky: .only

ml054 (commented):
@gregolsky I've fixed .only

PR #158, opened 2018-10-15
Enhance dates storage

gregolsky:
- store dates in utc
- store dates timezone info


gregolsky:
need to fix tests on travis

PR #217, opened 2020-04-09
fix exceptions on PutCompareExchangeValueOperation

cabelitos:
Hello everyone. We are having some problems on our project when trying to PutCompareExchangeValueOperation which contains some special characters.
Something like: "permissões" (it means permissions in portuguese).
It's important to note that if I create the key using the web studio, everything works fine.
However when using the nodejs API an exception is thrown.

So this commit adds the support for encoding the key property of
PutCompareExchangeValueOperation, thus allowing it to be used
with special characters which would break the URL otherwise.
This function is already used on all other operations, like
GetCompareExchange so it's safe to use in here too.

### DIFF
--- a/src/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.ts
+++ b/src/Documents/Operations/CompareExchange/PutCompareExchangeValueOperation.ts
@@ -72,7 +72,7 @@ export class PutCompareExchangeValueCommand<T> extends RavenCommand<CompareExcha
     }
 
     public createRequest(node: ServerNode): HttpRequestParameters {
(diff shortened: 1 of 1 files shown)


barbieri (approved):

alanpazetto (approved):

lgturatti (approved):

JanaMarinS (approved):

walterInacio (approved):

gregolsky (approved):
Thanks!
It looks good to me. I'll fix the Delete one in addition to this one.

cabelitos:
@gregolsky thank you!

gregolsky:
Landed on NPM 
image

PR #219, opened 2020-06-23
updated packages + request -> node-fetch

ml054:


package.json
```diff
@@ -52,48 +52,51 @@
...
+    "@types/verror": "^1.10.4",
+    "bluebird": "^3.7.2",
+    "change-case": "^3.1.0",
+    "deprecate": "^1.1.1",
+    "md5": "^2.2.1",
+    "moment": "^2.27.0",
+    "node-fetch": "^2.6.0",
+    "object.entries": "^1.1.2",
+    "object.values": "^1.1.1",
+    "pluralize": "^8.0.0",
```
gregolsky: Have you reviewed the changes for this one? It might change the behavior for collection name determination, leading to breaking changes right?

package.json
```diff
@@ -52,48 +52,51 @@
...
+    "qs": "^6.9.4",
+    "readable-stream": "^3.6.0",
+    "safe-memory-cache": "^2.0.0",
```
gregolsky, line 94: Please check what's been done there. We've been using some internal property to handle some special cases.

package.json
gregolsky: If it in fact is a breaking change, we should update that in upcoming 5.0.

ml054: reverting then

PR #242, opened 2020-12-02
sync with latest changes

ml054:


test/Ported/Server/Replication/PullReplicationTest.ts
```diff
@@ -328,7 +328,7 @@ import { PullReplicationDefinition } from "../../../../src/Documents/Operations/
         } finally {
             sink.dispose();
         }
-    });
+    }).timeout(60_000);
```
gregolsky: I remember doing `this.timeout(...)` inside the test func worked in the past. I think that's more explicit like this. What do you think?

gregolsky (approved):

test/Ported/Server/Replication/PullReplicationTest.ts
ml054: done

PR #274, opened 2021-04-18
RDBC-424

ml054:


src/Mapping/ObjectMapper.ts
```diff
@@ -372,7 +372,7 @@ export class TypesAwareObjectMapper implements ITypesAwareObjectMapper {
         }
 
         // tslint:disable-next-line:new-parens
-        return new (Function.prototype.bind.apply(ctor)) as TResult;
+        return new ctor() as TResult;
```
gregolsky, line 374: so that tslint rule disable is not applicable anymore, is it?

ml054: done

PR #282, opened 2021-07-09
using SIGKILL for stopping RavenDB instances after tests

ml054:


gregolsky:
Server not closing down on SIGTERM might be a sign of some bigger issue and we should investigate that separately. 
Could you create a ticket to check that please? You can assign it to me. Please cc Paweł there.

ml054:
https://issues.hibernatingrhinos.com/issue/RavenDB-16983

PR #289, opened 2021-11-19
RDBC-509 SaveChanges after query modifies the change vector

ayende:
- We need to avoid changing the metadataInstance if the changes we want to make are already there. This will cause us to save the document even though it wasnt changed.

### DIFF
--- a/src/Documents/Session/EntityToJson.ts
+++ b/src/Documents/Session/EntityToJson.ts
@@ -112,15 +112,16 @@ export class EntityToJson {
                 documentInfo.metadata[CONSTANTS.Documents.Metadata.RAVEN_JS_TYPE] || typeInfo.typeName;
         }
 
-        function differentNestedTypes() {
-            const existing =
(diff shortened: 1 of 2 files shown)


src/Documents/Session/EntityToJson.ts
```diff
@@ -112,10 +112,29 @@ export class EntityToJson {
...
+                }
+            }
+            return false;
+        }
+
         if (documentInfo.metadataInstance) {
-            documentInfo.metadataInstance[CONSTANTS.Documents.Metadata.NESTED_OBJECT_TYPES] = typeInfo.nestedTypes;
-            documentInfo.metadataInstance[CONSTANTS.Documents.Metadata.RAVEN_JS_TYPE] =
-               documentInfo.metadataInstance["Raven-Node-Type"] || typeInfo.typeName;
+            if(differentNestedTypes()){
```
ayende: The problem is that here it returns a `Proxy`, so we can't do an easy comparison. I decided to just handle that explicitly, but not sure if this is the best way.

ml054 (approved):

PR #297, opened 2022-02-09
RDBC-543 Serialization issue ( for objects w/o proto )

ml054:


gregolsky:
Maybe it's a silly question, but.. Is it somehow related to this https://github.com/ravendb/ravendb-nodejs-client#using-object-literals-for-entities ? I think those objs do not have prototype as well.

image

ml054:
image
@gregolsky No, this is different story.

PR #307, opened 2022-05-01
Stats invalid when performing aggregate query and normal query lazily

nickchampion:
Failing test for this issue

https://github.com/ravendb/ravendb/issues/14149


ml054:
Please have a look at: https://github.com/ravendb/ravendb/issues/14149#issuecomment-1114639743

PR #316, opened 2022-07-06
Correct chunk joining to attribute

configurator:
Array.join() implicitly converts the array items to strings, while we need to join the code points before converting to string.

This means when a character is split between two chunks, both of its parts are converted to unicode replacement characters.

This change fixes that.


gregolsky:
Tests failures seem unrelated. Periodic ones run on master show that the server behavior changed.
image

ayende:
Handled by #318 with a test + more fixes

PR #320, opened 2022-07-27
Prevent hash collisions

configurator:
Fixes two collisions:
- different numbers map to the same hash if equal modulo 256
- different types hash identically; `1` and `'1'` are both equal to `true`, for example


ml054:
@configurator Thanks for your contribution. Nice catch.

ml054:
Fixed in 5.2.6

PR #326, opened 2022-09-14
Drop support for node < 14.0.0

nemanja-tosic:
Support actively maintained LTS versions only.


ml054:
opened separate PR for 5.4: https://github.com/ravendb/ravendb-nodejs-client/pull/332

Thanks @nemanja-tosic

PR #339, opened 2022-11-25
RDBC-647 small fixes + fix links

Danielle9897:
* Small fixes
* Fix headings size
* Fix links


Danielle9897:
@ml054 links are now fixed with correct sha1

ml054 (approved):

PR #340, opened 2022-11-27
RDBC-647 Add intro

Danielle9897:
@ml054
IMHO - this intro will be useful

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,24 +16,24 @@ npm install --save ravendb
 ## Documentation
 
 * This readme provides short examples for the following:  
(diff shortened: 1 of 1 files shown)


ml054 (changes requested):
@Danielle9897 Could you check if it will work if you strip https://github.com/ravendb/ravendb-nodejs-client from each url? 

I'd like to avoid hardcoding fork name in urls

Danielle9897:
@ml054 
I made links 'local' to the page.

ml054:
Perfect, thanks. :)

PR #343, opened 2022-12-07
RDBC-658 AddDatabaseNodeOperation: Don't add node param to request wh…

Danielle9897:
…en node is undefined

### DIFF
--- a/src/ServerWide/Operations/AddDatabaseNodeOperation.ts
+++ b/src/ServerWide/Operations/AddDatabaseNodeOperation.ts
@@ -11,11 +11,11 @@ import { RaftIdGenerator } from "../../Utility/RaftIdGenerator";
 
 export class AddDatabaseNodeOperation implements IServerOperation<DatabasePutResult> {
     private readonly _databaseName: string;
-    private readonly _node: string;
+    private readonly _nodeTag: string;
 
     public constructor(databaseName: string, node?: string) {
         this._databaseName = databaseName;
-        this._node = node;
+        this._nodeTag = node;
     }
 
     public get resultType(): OperationResultType {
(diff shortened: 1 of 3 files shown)


src/ServerWide/Operations/AddDatabaseNodeOperation.ts
```diff
@@ -45,7 +45,7 @@ class AddDatabaseNodeCommand extends RavenCommand<DatabasePutResult> implements
     createRequest(node: ServerNode): HttpRequestParameters {
         let uri = node.url + "/admin/databases/node?name=" + this._databaseName;
 
-        if (node) {
+        if (node && this._node) {
```
ml054: why we need `if (node` check ?

ml054: If node is null/undefined it will throw any way in line 46

Danielle9897: * fixed
* refactored` this._node` to `this._nodeTag` (I hope that is ok)

ml054 (approved):

Danielle9897:
Using the env-var worked great
Tests pass for me locally

ml054:
I knew it will help. I caught myself so many times on that. :D

test/Issues/RDBC_658.ts
```diff
@@ -0,0 +1,82 @@
...
+    let testContext: ClusterTestContext;
+
+    beforeEach(async () => testContext = new ClusterTestContext());
+    afterEach(async () => testContext.dispose());
+
+    describe("Add database node", function () {
+
+        it("Add random database node", async () => {
+            try {
+                testContext = new ClusterTestContext();
```
ml054: please compare this test and any other with cluster test context.

https://github.com/ravendb/ravendb-nodejs-client/blob/7db833e14a10c92cab91ff8b9276deb7744e5f17/test/Ported/Issues/RDBC_538.ts#L6

IE. we don't override textContext variable in test

I bet that's the reason tests fails

PR #344, opened 2022-12-16
Refactor createEmptyObject as a protected method to allow customization

nemanja-tosic:
See: https://issues.hibernatingrhinos.com/issue/RDBC-661/NodeJS-client-should-not-invoke-constructor-when-mapping-entities-from-object-literals

Note: i've made the `createEmptyObject` method protected as we'd like to be able to extend the TypesAwareObjectMapper behavior at certain points. Currently, it only has the interface public methods, and everything else are private methods which prevents us from changing small bits unless we ignore typescript errors. We have a couple of things we'd like to contribute so we'll probably make incremental suggestions.

### DIFF
--- a/src/Mapping/ObjectMapper.ts
+++ b/src/Mapping/ObjectMapper.ts
@@ -1,4 +1,4 @@
-import { ObjectTypeDescriptor, ClassConstructor, ObjectLiteralDescriptor, EntityConstructor } from "../Types";
+import { ObjectTypeDescriptor, ObjectLiteralDescriptor, EntityConstructor } from "../Types";
 import { throwError } from "../Exceptions";
 import { TypeUtil } from "../Utility/TypeUtil";
 import { getLogger } from "../Utility/LogUtil";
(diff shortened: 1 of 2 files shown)


nemanja-tosic:
`Object.create(Ctor.Prototype)` is not equivalent to `new Ctor()` without a constructor call since the addition of private class fields. In order for private class fields to work, a class MUST be initialized with new. We had issues with getters/setters accessing a private class fields when the constructor was not invoked.

For the time being, we will not use private class fields, but will use typescript's private modifier alongside the "no constructor invocation" approach initially proposed in the PR. I still think it's a good thing to expose `createEmptyObject` as a protected factory method so that clients have flexibility when instantiating objects.

ml054:
Hi @nemanja-tosic 

Sorry for long response time. 

changing from calling ctr to not calling it it breaking change. 

I think we can implement that in following way:
- add new property to document conventions which allows user to select object creation strategy - either `Object.create` or `new ctr()`
-  default value should be `new ctr()`
- in 6.0 we might change default behavior to `Object.create`

What do you think ?

nemanja-tosic:
Unfortunately, given the development of ECMAScript regarding class fields, using Object.create won't work because it does not initialize class fields. As far as i know (transpilers aside), fields are initialized as part of the `new` call, either using `new Class()` or `Reflect.construct` (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new) so it would appear that the mapper must call `new ctor()` as it does now in order to properly initialize fields. From this point i have some ideas, but the implementation will depend on factors such as whether or not transpilers are used.

For this PR, i'd only like to expose the `createEmptyObject` method as a protected factory method and allow client code to modify the instantiation by subclassing `TypesAwareObjectMapper`. I'll try to make these changes today, they should be backwards compatible and i'll reopen the PR then.

ml054:
@nemanja-tosic Thanks for your contribution.

PR #345, opened 2022-12-17
RDBC-662 store type of every array member

nemanja-tosic:
The change introduces a different way to store array field information, but is backwards compatible with the previous version.

The PR relies on the same mechanism as objects for type information.

### DIFF
--- a/src/Mapping/ObjectMapper.ts
+++ b/src/Mapping/ObjectMapper.ts
@@ -425,8 +425,7 @@ export class TypesAwareObjectMapper implements ITypesAwareObjectMapper {
         }
 
         if (Array.isArray(obj)) {
-            const newObjPathPrefix = `${objPathPrefix}[]`;
-            return obj.map(x => this._makeObjectLiteral(x, newObjPathPrefix, typeInfoCallback, knownTypes));
+            return obj.map((x, index) => this._makeObjectLiteral(x, `${objPathPrefix}.${index}`, typeInfoCallback, knownTypes));
         }
 
         if (TypeUtil.isObject(obj)) {
--- a/test/Mapping/ObjectMapperTests.ts
+++ b/test/Mapping/ObjectMapperTests.ts
@@ -134,6 +134,36 @@ describe("ObjectMapper", function () {
                 assert.strictEqual(d.getMonth(), 4);
             }
         });
+        
+        it("can handle polymorphic arrays", () => {
+            class Comedy extends Movie {}
+            
+            const typeInfo = {
+                nestedTypes: {
+                    "movies.0": "Movie",
+                    "movies.1": "Comedy"
+                }
+            };
+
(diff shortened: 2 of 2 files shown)


nemanja-tosic:
@ml054 does this look like something we could merge?

ml054:
Hi, @nemanja-tosic I just returned from vacations. Will try to handle your PRs in next week. Sorry for delay.

ml054:
Again, I'm really sorry for delay. 

I've seen your changes. Due to possibility of breaking changes and increase of entries in nested types, I think you might merge after making one small adjustment:

In `TypesAwareObjectMapper` we have reference to `DocumentConventions`. 

In `DocumentConventions` we might introduce new configuration option to configure if we should store types for each array member (and allow polimorphic arrays) or store single type (which is current approach).

nemanja-tosic:
I'm curious what does the C# client for example do in this instance?

The way the client handles things now should be considered a bug in my mind because what i store is not what i get back on hydration. There are no errors, so it bit us a couple of times.

ml054:
```
public class Animal
    {
        public string Name { get; set; }
    }

    public class Dog : Animal
    {
        public string DogField { get; set; }
    }

    public class Cat2 : Animal
    {
        public string CatField { get; set; }
    }
    
    public class Container
    {
        public Animal[] Animals { get; set; }
    }
```

Test:
```
 using (IDocumentSession session = store.OpenSession())
                {
                    var dog = new Dog {Name = "d1", DogField = "dd"};

                    var cat = new Cat2 {Name = "cat name", CatField = "cc"};
                    
                    session.Store(new Container
                    {
                        Animals = new Animal[] { dog, cat }
                    });
                    session.SaveChanges();
                    
                    WaitForUserToContinueTheTest(store);
                }
```

Saved document:
```
{
    "Animals": [
        {
            "$type": "SlowTests.MailingList.Dog, SlowTests",
            "DogField": "dd",
            "Name": "d1"
        },
        {
            "$type": "SlowTests.MailingList.Cat2, SlowTests",
            "CatField": "cc",
            "Name": "cat name"
        }
    ],
    "@metadata": {
        "@collection": "Containers",
        "Raven-Clr-Type": "SlowTests.MailingList.Container, SlowTests"
    }
}
```

nemanja-tosic:
Looking at this, i feel that nodejs client has behavior that would not be expected seeing how it doesn't support polymorphic arrays, which should be considered a bug. Using conventions to support unexpected behavior doesn't feel right but i am open to adding that to the PR i suppose.

ml054:
Ok I agree with you. Could you add test which shows that this change is backward compatible and can properly load data stored in old format?

nemanja-tosic:
The test suite for `fromObjectLiteral` was unchanged, so the changes should be backwards compatible. If we need to explicitly note that this is a "legacy" format, we could do that.

ml054:
Merged. Thanks.

PR #353, opened 2023-02-05
Fix payload mapping for mentorNode in ETL configuration

nickchampion:
I believe this mapping is wrong, the studio uses `MentorNode` not `MentorName` and when I create / update an ETL task via the node client the responsible node for the task is not set correctly.

Also not sure whether SubscriptionState should have a `mentorNode` instead of `mentorName` field, let me know and I can update the PR

`src/Documents/Subscriptions/SubscriptionState.ts`

```
export interface SubscriptionState {
    query: string;
    changeVectorForNextBatchStartingPoint: string;
    subscriptionId: number;
    subscriptionName: string;
    mentorName: string;
    nodeTag: string;
    lastBatchAckTime: string;
    lastClientConnectionTime: string;
    disabled: boolean;
}

```

### DIFF
--- a/src/Documents/Operations/Etl/EtlConfiguration.ts
+++ b/src/Documents/Operations/Etl/EtlConfiguration.ts
@@ -15,7 +15,7 @@ export class EtlConfiguration<T extends ConnectionString> {
         return {
             TaskId: this.taskId,
             Name: this.name,
-            MentorName: this.mentorNode,
+            MentorNode: this.mentorNode,
             ConnectionStringName: this.connectionStringName,
             Transforms: this.transforms.map(x => serializeTransformation(x)),
             Disabled: this.disabled,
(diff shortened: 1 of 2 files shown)


ml054:
Hi @nickchampion 

Thanks for catching this. 

You are absolutely right, second place should be changed as well. It is c# version: https://github.com/ravendb/ravendb/blob/f5f77ef8c11b84a0d90de13b7c4edfcb600affd0/src/Raven.Client/Documents/Subscriptions/SubscriptionState.cs#L20

nickchampion:
No problem @ml054 I've updated the PR with the change. Thanks

ml054:
Thanks Running tests and we are ready for merge.

ml054 (approved):

ml054:
@nickchampion Merged. Thanks for your contribution.

PR #357, opened 2023-03-03
RDBC-681 @nested-object-types doesn’t store type information for newly added f…

ml054:
…ields on update


ml054:
https://issues.hibernatingrhinos.com/issue/RDBC-681/nested-object-types-doesnt-store-type-information-for-newly-added-fields-on-update

PR #360, opened 2023-03-18
RDBC-654 Investigate cloudflare and node.js client interoperability

ml054:


src/Http/RequestExecutor.ts
```diff
@@ -399,14 +423,25 @@ export class RequestExecutor implements IDisposable {
...
         executor._firstTopologyUpdatePromise.catch(TypeUtil.NOOP);
 
         return executor;
     }
 
+    private static getGlobalApplicationIdentifier() {
+        // due to cloudflare constraints we can't init GLOBAL_APPLICATION_IDENTIFIER in static
+
+        if (!this.GLOBAL_APPLICATION_IDENTIFIER) {
+            this.GLOBAL_APPLICATION_IDENTIFIER = uuidv4();
```
ayende, line 439: Do we need `.toString()` here? It used to be there, and the type changed?
That would be we send `[Object object]` ?

ml054: uuid4 seems to have toString():
image

PR #418, opened 2024-03-29
RDBC-432 get rid of bluebird from node.js client

ml054:


src/Documents/DocumentStore.ts
```diff
@@ -112,21 +112,24 @@ export class DocumentStore extends DocumentStoreBase {
...
 
         disposeChain
-            .then(() => {
+            .then(async () => {
                 if (this._multiDbHiLo) {
-                    return Promise.resolve()
-                        .then(() => this._multiDbHiLo.returnUnusedRange())
-                        .catch(err => this._log.warn("Error returning unused ID range.", err));
+                    try {
+                        await Promise.resolve();
```
gregolsky, line 121: this line can be removed

PR #443, opened 2024-08-05
RDBC-866 Fix node.js tsdocs + add debug symbols

ml054:


src/Documents/BulkInsertOperation.ts
```diff
@@ -553,6 +553,19 @@ export class BulkInsertOperation extends BulkInsertOperationBase<object> {
         return { id, metadata, getId };
     }
 
+    /**
+     * Synchronous version of store. In stores entity unless buffer is full (then false is returned).
```
ayende:
```suggestion
     * Synchronous version of store. It stores the entity unless the buffer is full (then false is returned and you are expected to call await store() ).
```

ml054: thanks

PR #452, opened 2024-08-10
V6: Ignore temp directory .tshy-build/

ngbrown:
The `tshy` command creates a bunch of files in the `.tshy-build/` directory, even if temporarily, they should be ignored by git.


ml054:
@ngbrown Thanks for contribution.

Happy hacking with v6.0 version. :)

PR #476, opened 2025-04-24
RDBC-903 - Drop Node.js 18 support

M4xymm:
Increase minimum Node.js version to 20.0.0 and update GitHub Actions workflow to use v7.0 branch. Remove support for Node.js 18.x in the test matrix to align with updated requirements.


kalczur:
update version in package-lock.json

kalczur (approved):

PR #481, opened 2025-07-02
RDBC-909 - Returning class instances in client causes issues in Next.js — responses should be plain objects

M4xymm:
Introduce `returnPlainJsObjects` convention to optionally convert API results to plain objects. Update query methods to honor this setting by utilizing `JsonSerializer.toPlainObject`. Include support for environments requiring serializable results, such as Next.js.

### DIFF
--- a/src/Documents/Session/AbstractDocumentQuery.ts
+++ b/src/Documents/Session/AbstractDocumentQuery.ts
@@ -97,6 +97,7 @@ import {
 } from "./IVectorFieldFactory.js";
 import { VectorEmbeddingFieldFactory } from "../Queries/VectorSearch/VectorEmbeddingFieldFactory.js";
 import { Field } from "../../Types/index.js";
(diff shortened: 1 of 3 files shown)


src/Documents/Session/AbstractDocumentQuery.ts
```diff
@@ -2193,7 +2194,13 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
 
     public async all(): Promise<T[]> {
         const results = await this.iterator();
-        return [...results];
+        const array = [...results];
+
+        if (this.conventions.returnPlainJsObjects) {
+            return array.map(item => JsonSerializer.toPlainObject(item))
```
kalczur, line 2200: Can item be `undefined`? if yes we will get SyntaxError.

M4xymm: Done.

kalczur (approved):

PR #482, opened 2025-07-07
RDBC-908 Node.js Client malfunction in ssr applications

M4xymm:
The issue occurs in server components when using client versions above 6.0

### DIFF
--- a/src/Http/RequestExecutor.ts
+++ b/src/Http/RequestExecutor.ts
@@ -378,28 +378,28 @@ export class RequestExecutor implements IDisposable {
         }
     }
 
-private static async createAgent(options: Agent.Options) {
-    try {
-        let AgentInstance;
-
+    private static async createAgent(options: Agent.Options) {
         try {
-            const undiciModule = await import(importFix("undici"));
-            AgentInstance = undiciModule.Agent;
+            let UndiciAgent;
+
+            try {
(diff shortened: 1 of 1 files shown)


src/Http/RequestExecutor.ts
```diff
@@ -378,15 +378,28 @@ export class RequestExecutor implements IDisposable {
         }
     }
 
-    private static async createAgent(options: Agent.Options) {
+private static async createAgent(options: Agent.Options) {
+    try {
+        let AgentInstance;
```
kalczur: It's not instance. Rename it to just `Agent`

kalczur (changes requested):

src/Http/RequestExecutor.ts
```diff
@@ -378,15 +378,28 @@ export class RequestExecutor implements IDisposable {
         }
     }
 
-    private static async createAgent(options: Agent.Options) {
+private static async createAgent(options: Agent.Options) {
```
kalczur: missing tab

src/Http/RequestExecutor.ts
M4xymm: Renamed to UndiciAgent

kalczur (approved):

PR #505, opened 2025-12-10
RDBC-969 Query with whereIn using the exact parameter doesn't work

### LINKED ISSUE #500: Query with whereIn using the exact parameter doesn't work
I am trying to do a case sensitive query using `whereIn` but the `exact` parameter doesn't seem to do anything, I get the same results.

as an example:
```typescript
const query = session.query({ indexName: `${tableId}-index` });
query.whereIn(externalId, selectedMarkerIds, true); // true here is supposed to make it a case sensitive match
```

Looking at the code, I can see that there is a potentially missing `WhereOptions` parameter that should be passed through to the `WhereToken.create` function:

```typescript
// src/Documents/Session/AbstractDocumentQuery.ts

    public _whereIn(fieldName: string, values: any[], exact: boolean = false): void {
        this._assertMethodIsCurrentlySupported("whereIn");

        fieldName = this._ensureValidFieldName(fieldName, false);

        const tokens = this._getCurrentWhereTokens();
        this._appendOperatorIfNeeded(tokens);
        this._negateIfNeeded(tokens, fieldName);

        const whereToken = WhereToken.create(
            "In",
            fieldName,
            this._addQueryParameter(
                this._transformCollection(fieldName, AbstractDocumentQuery._unpackCollection(values))));
            // should add new WhereOptions({ exact })); as done in other functions?
        tokens.push(whereToken);
    }
```
but adding that `WhereOptions({ exact })` parameter doesn't fix the problem.

here is a test I added that shows the issue:
```typescript
// test/Ported/QueryTest.ts
it("query with where in exact parameter", async () => {
    const session = store.openSession();
    const query = session.query(User)
        .whereIn("name", ["tarzan", "no_such"], true);

    const queryString = query.toString();

    const results = await query.all();
    assert.strictEqual(results.length, 0);

    // Verify that the exact parameter is properly passed through to the query
    assertThat(queryString)
        .contains("exact(");
});

```
  
and you can see the results:
```
QueryTest
    with regular users set
DEBUG: RavenDB server URL http://127.0.0.1:58901
      1) query with where in exact parameter
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal: 1 !== 0

    at Context.<anonymous> (file:///Users/simon.groom/Code/ravendb-nodejs-client/test/Ported/QueryTest.ts:218:20)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: '1',
  expected: '0',
  operator: 'strictEqual'
}


  0 passing (2s)
  1 failing

  1) QueryTest
       with regular users set
         query with where in exact parameter:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal: 1 !== 0

      + expected - actual

      -1
      +0
      
      at Context.<anonymous> (file:///Users/simon.groom/Code/ravendb-nodejs-client/test/Ported/QueryTest.ts:218:20)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

I have a workaround in that I can just to a raw RQL query and this does what I want.

M4xymm:
Fixes #500

### DIFF
--- a/src/Documents/Session/AbstractDocumentQuery.ts
+++ b/src/Documents/Session/AbstractDocumentQuery.ts
@@ -1054,7 +1054,10 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
             "In",
             fieldName,
             this._addQueryParameter(
-                this._transformCollection(fieldName, AbstractDocumentQuery._unpackCollection(values))));
+                this._transformCollection(fieldName, AbstractDocumentQuery._unpackCollection(values))
+            ),
+            new WhereOptions({exact})
+        );
         tokens.push(whereToken);
     }
 
--- a/test/Ported/QueryTest.ts
+++ b/test/Ported/QueryTest.ts
@@ -264,6 +264,24 @@ describe("QueryTest", function () {
             assert.strictEqual(results.length, 1);
         });
 
+        // RavenDB-25553 - dynamic queries with exact search in whereIn clause are not working as expected
+        it.skip("query with where in and exact", async () => {
+            const session = store.openSession();
+
+            const user4 = Object.assign(new User(), {name: "tarzan"});
+
+            await session.store(user4, "users/4");
+            await session.saveChanges();
+
+            const resultsExactTarzan = await session.query(User).whereIn("name", ["Tarzan"], true).all()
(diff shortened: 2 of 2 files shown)


M4xymm:
image

PR #539, opened 2026-03-26
RDBC-1050: Fix critical and high-severity bugs found in systematic search on NodeJS client

redknightlois:
### Issue link
https://issues.hibernatingrhinos.com/issue/RDBC-1050

### Additional description

Fixes 6 verified bugs (3 critical, 3 high) discovered during a systematic gap analysis comparing the NodeJS client against the C# reference client on the v7.2 branch. Each bug was confirmed with a failing test before the fix was applied.

  1. `BulkInsertOperation._handleErrors` calls async `_getExceptionFromOperation()` without await, throwing a raw Promise object instead of the actual server error
  2. `VectorOptions.fromJson` scrambles Dimensions/NumberOfEdges/NumberOfCandidates into wrong constructor parameters; `dimensions` field missing from class
  3. `Highlightings.resultIndents` uses `Object.keys()` on a `Map`, always returning `[]`
  4. `LazyGetCompareExchangeValuesOperation` has inverted `startsWith` condition, never sending the prefix to the server
  5. `DeletedEntitiesHolder.prepareEntitiesDeletes` dispose uses regular function where `this` binds to the object literal, not the holder instance
  6. `DocumentsById.getTrackedEntities` uses bracket notation on a `Map`, bypassing `Map.set()` so entries are invisible to `Map.get()`/`Map.has()`/`Map.size`

### Type of change

  - [x] Bug fix
  - [ ] Regression bug fix
  - [ ] Optimization
  - [ ] New feature

### How risky is the change?

  - [x] Low
  - [ ] Moderate
  - [ ] High
  - [ ] Not relevant

### Backward compatibility

  - [x] Non breaking change
  - [ ] Breaking change
  - [ ] Not relevant

### Is it platform specific issue?

  - [ ] Yes. Please list the affected platforms.
  - [x] No

### Documentation update

  - [ ] This change requires a documentation update. Please mark the issue on YouTrack using `Documentation Required` tag.
  - [x] No documentation update is needed

### Testing by Contributor

  - [ ] Internal classes added to the test class (e.g. entity or index definition classes) have the lowest possible access modifier (preferable `private`)
  - [ ] It has been verified by manual testing

### Testing by RavenDB QA team

  - [ ] This change requires a special QA testing due to possible performance or resources usage implications (CPU, memory, IO). Please mark the issue on YouTrack using `QA Required`
  tag.
  - [x] No special testing by RavenDB QA team is needed

  - [ ] Yes. Please list the affected features/subsystems and provide appropriate explanation
  - [x] No

### UI work

  - [ ] It requires further work in the Studio. Please mark the issue on YouTrack using `Studio Required` tag.
  - [x] No UI work is needed


src/Documents/Session/Operations/Lazy/LazyGetCompareExchangeValuesOperation.ts
```diff
@@ -101,7 +101,7 @@ export class LazyGetCompareExchangeValuesOperation<T> implements ILazyOperation
         } else {
             pathBuilder = new StringBuilder("?");
 
-            if (StringUtil.isNullOrEmpty(this._startsWith)) {
+            if (StringUtil.isNullOrEmpty(this._startsWith) == false) {
```
M4xymm:
```suggestion
            if (!StringUtil.isNullOrEmpty(this._startsWith)) {
```

M4xymm (changes requested):

src/Documents/Session/Operations/Lazy/LazyGetCompareExchangeValuesOperation.ts
redknightlois: I make that change because we use the convention of instead of using ! we use == false in the whole server and C# code. Is the convention here different?

M4xymm: Hi, Yes. At node, we use this convention because it is simply "good practice" used, in fact, by a much larger group of js/ts programmers. ``== false`` is typically C#. :)

M4xymm (approved):

PR #541, opened 2026-04-07
RDBC-1053: Implemented several fixes.

redknightlois:
### Issue link

https://issues.hibernatingrhinos.com/issue/RDBC-1053

### Additional description

  - **BatchPatchCommandData**: duplicate document IDs silently accepted instead of throwing
  - **AbstractDocumentQuery.closeSubclause**: nested subclause boost iteration doesn't stop after finding the matching open token
  - **JsonOperation**: field type change from scalar to array produces spurious array change records alongside the correct FieldChanged
  - **AbstractDocumentQuery range queries**: `whereBetween`, `whereGreaterThan`, `whereLessThan`, `whereGreaterThanOrEqual`, `whereLessThanOrEqual` all treat `0`, `""`, and `false` as
  null/unbounded
  - **AbstractDocumentQuery.isFilterActive**: checks bottom of filter mode stack instead of top, breaking nested filter scopes
  - **DocumentConventions.identityPartsSeparator**: setter validates the current value instead of the incoming value, allowing `"|"` on first set and blocking all subsequent changes
  - **TimeValue.compareTo**: cross-unit comparison (months vs seconds) compares the same value's bounds against itself instead of the other value's bounds
  - **TimeValue._isSpecialCompare**: `MIN_VALUE.compareTo(MIN_VALUE)` returns `-1` instead of `0` due to swapped min/max checks
  - **AbstractDocumentQuery._fuzzy/_proximity**: null token list causes `TypeError` instead of the intended error message

### Type of change

- [x] Bug fix
- [ ] Regression bug fix
- [ ] Optimization
- [ ] New feature

### How risky is the change?

- [x] Low
- [ ] Moderate
- [ ] High
- [ ] Not relevant

### Backward compatibility

- [x] Non breaking change
- [ ] Breaking change
- [ ] Not relevant

### Is it platform specific issue?

- [ ] Yes. Please list the affected platforms.
- [x] No

### Documentation update

- [ ] This change requires a documentation update. Please mark the issue on YouTrack using `Documentation Required` tag.
- [ ] No documentation update is needed

### Testing by Contributor

- [ ] Internal classes added to the test class (e.g. entity or index definition classes) have the lowest possible access modifier (preferable `private`)
- [ ] It has been verified by manual testing

### Testing by RavenDB QA team

- [ ] This change requires a special QA testing due to possible performance or resources usage implications (CPU, memory, IO). Please mark the issue on YouTrack using `QA Required` tag.
- [x] No special testing by RavenDB QA team is needed

- [ ] Yes. Please list the affected features/subsystems and provide appropriate explanation
- [x] No

### UI work

- [ ] It requires further work in the Studio. Please mark the issue on YouTrack using `Studio Required` tag.
- [x] No UI work is needed

### DIFF
--- a/src/Documents/Session/AbstractDocumentQuery.ts
+++ b/src/Documents/Session/AbstractDocumentQuery.ts
@@ -213,7 +213,7 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
     protected _explanationToken: ExplanationToken;
 
     protected isFilterActive(): boolean {
-        return this._filterModeStack.length && this._filterModeStack[0];
+        return this._filterModeStack.length && this._filterModeStack.at(-1);
     }
 
 
@@ -1139,9 +1139,9 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
         endParams.fieldName = fieldName;
 
         const fromParameterName = this._addQueryParameter(
-            !start ? "*" : this._transformValue(startParams, true));
+            start == null ? "*" : this._transformValue(startParams, true));
         const toParameterName = this._addQueryParameter(
-            !end ? "NULL" : this._transformValue(endParams, true));
+            end == null ? "NULL" : this._transformValue(endParams, true));
 
         const whereToken = WhereToken.create(
             "Between", fieldName, null, new WhereOptions({
@@ -1174,7 +1174,7 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
         whereParams.fieldName = fieldName;
 
         const parameter = this._addQueryParameter(
-            !value ? "*" : this._transformValue(whereParams, true));
+            value == null ? "*" : this._transformValue(whereParams, true));
 
         const whereToken = WhereToken.create(
             "GreaterThan", fieldName, parameter, new WhereOptions({ exact }));
@@ -1203,7 +1203,7 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
         whereParams.fieldName = fieldName;
 
         const parameter = this._addQueryParameter(
-            !value ? "*" : this._transformValue(whereParams, true));
+            value == null ? "*" : this._transformValue(whereParams, true));
         const whereToken = WhereToken.create(
             "GreaterThanOrEqual", fieldName, parameter, new WhereOptions({ exact }));
         tokens.push(whereToken);
@@ -1223,7 +1223,7 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
         whereParams.fieldName = fieldName;
 
         const parameter = this._addQueryParameter(
-            !value ? "NULL" : this._transformValue(whereParams, true));
+            value == null ? "NULL" : this._transformValue(whereParams, true));
         const whereToken = WhereToken.create(
             "LessThan", fieldName, parameter, new WhereOptions({ exact }));
         tokens.push(whereToken);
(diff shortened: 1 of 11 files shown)


src/Documents/Session/AbstractDocumentQuery.ts
```diff
@@ -213,7 +213,7 @@ export abstract class AbstractDocumentQuery<T extends object, TSelf extends Abst
     protected _explanationToken: ExplanationToken;
 
     protected isFilterActive(): boolean {
-        return this._filterModeStack.length && this._filterModeStack[0];
+        return this._filterModeStack.length && this._filterModeStack[this._filterModeStack.length - 1];
```
M4xymm: This snippet can be written shorter, but it does not change the result.

```suggestion
        return this._filterModeStack.length && this._filterModeStack.at(-1);
```

test/Documents/Session/AbstractDocumentQueryTest.ts
```diff
@@ -0,0 +1,90 @@
...
+    afterEach(() => {
+        store?.dispose();
+    });
+
+    function createStore() {
+        store = new DocumentStore("http://localhost:9999", "TestDB");
+        store.conventions.findCollectionNameForObjectLiteral = () => "test";
+        store.initialize();
+        return store;
+    }
```
M4xymm: example of testing with DocumentStore

https://github.com/ravendb/ravendb-nodejs-client/blob/c5d5e9805f94e2c829dd1268da163a7db8942141/test/Ported/Documents/Operations/GenAiEtlTest.ts#L15-L22

```suggestion
    let store: IDocumentStore;

    beforeEach(async function () {
        store = await testContext.getDocumentStore();
    });

    afterEach(async () =>
        await disposeTestDocumentStore(store));
```

test/Documents/Session/AbstractDocumentQueryTest.ts
```diff
@@ -0,0 +1,90 @@
...
+    it("whereBetween preserves 0 as a valid boundary value", function () {
+        const s = createStore().openSession();
+
+        const q = s.query({ collection: "users" }).whereBetween("age", 0, 100);
+        const iq = q.getIndexQuery();
+
+        assert.strictEqual(iq.queryParameters.p0, 0,
+            "start value 0 should be preserved, not replaced with wildcard '*'");
+        assert.strictEqual(iq.queryParameters.p1, 100);
+    });
```
M4xymm, line 26: I know these are just tests, but it would be better to avoid abbreviations like `q` or `iq`, as they result in poor naming. Variable names should be descriptive enough that you can understand their purpose at a glance :)

```suggestion
    it("whereBetween preserves 0 as a valid boundary value", function () {
        const query = store.query({ collection: "users" }).whereBetween("age", 0, 100);
        const indexQuery = query.getIndexQuery();

        assert.strictEqual(indexQuery.queryParameters.p0, 0,
            "start value 0 should be preserved, not replaced with wildcard '*'");
        assert.strictEqual(indexQuery.queryParameters.p1, 100);
    });
```

M4xymm (changes requested):

test/Mapping/JsonOperationTest.ts
```diff
@@ -0,0 +1,24 @@
+import assert from "node:assert";
+import { describe, it } from "node:test";
```
redknightlois, line 2: @M4xymm this is because I cannot make it work with older mocha... maybe we should upgrade first. https://issues.hibernatingrhinos.com/issue/RDBC-1055/Upgrade-mocha-from-10.7.3-to-11.x

M4xymm (approved):

PR #573, opened 2026-08-17
[hugin] Port the 7.2.3 → 7.2.5 client surface

redknightlois:
This ports what the .NET client (`src/Raven.Client`) gained between 7.2.3 and 7.2.5 into the Node.js client, following this client's own conventions. Only the 7.2.3 to 7.2.5 changes are in scope: things the .NET client already had before 7.2.3 that were never ported are left alone, and anything the .NET client marks internal stays out of the public API.

One commit per feature, in dependency order; each one builds and passes its tests on its own:

- CDC Sink tasks: `AddCdcSinkOperation`, `UpdateCdcSinkOperation`, the configuration and column mapping types, and the task wired into ongoing tasks, smuggler and the database record
- Server-wide connection strings: GET, PUT and DELETE server operations and the `ServerWideConnectionString` type
- Reading AI conversation messages: `GetConversationMessagesOperation` with paging, detail levels and a null result on 404; every conversation run now sends `cancelPendingActionTools=false`, as the .NET client does
- Azure Service Bus in the Queue family: broker type, connection settings and sink source helpers
- Hub and sink cursors on the pull replication sink task info
- `disableChecksumValidation` on the S3 settings
- SSO members on certificate metadata: `usage`, `ssoServerPublicKeyPinningHashes`, `allowAnySsoServer`, `ssoIdentifiers`
- Certificate edit path: `EditClientCertificateOperation` takes `disabled` and always sends `Disabled`; the SSO fields are sent only when given
- `CLIENT_VERSION` pinned to 7.2.5; the npm package version in package.json stays 7.2.6

Checks: `tsc --noEmit` and eslint are clean; request and response shapes were checked against a live 7.2.5 server and against the .NET types; there is one test file per feature under `test/Ported`, and the ones that need a server are gated on the test server.

This is a new series that replaces the first one after the review. The comment below lists what changed and why some things were left out.

Produced by hugin's reforge workflow; each commit carries a `Reforge-Run` trailer.


src/ServerWide/Operations/Certificates/CertificateMetadata.ts
```diff
@@ -11,4 +31,8 @@ export interface CertificateMetadata {
     collectionSecondaryKeys?: string[];
     collectionPrimaryKey?: string;
     publicKeyPinningHash: string;
```
M4xymm, line 15: disabled is also missing from the metadata interface, while C# CertificateMetadata has public bool Disabled and writes it in ToJson. Reading certificate metadata will drop this field

src/ServerWide/Operations/Certificates/EditClientCertificateOperation.ts
```diff
@@ -78,13 +100,39 @@ class EditClientCertificateCommand extends RavenCommand<void> implements IRaftCo
         // serializer) so the casing-preserving serializer can be used: the default one would also
         // PascalCase the first letter of every `permissions` key (database names), breaking
         // case-sensitive permission matching (RDBC-1085).
-        const definition = {
+        const definition: any = {
             Thumbprint: this._thumbprint,
```
M4xymm, line 104: the definition payload omits Disabled, and the operation has no disabled parameter at all. The C# command always serializes Disabled (EditClientCertificateOperation.cs:119), so editing a disabled certificate through this client would silently re-enable it.

src/Documents/Operations/Etl/ConnectionString.ts
```diff
@@ -14,16 +15,48 @@ export type ConnectionStringType =
     | "Olap"
     | "ElasticSearch"
     | "Queue"
+    | "Snowflake"
```
M4xymm: "Snowflake" and "SnowflakeEtl" were added to the type unions, but there is no SnowflakeConnectionString class or SnowflakeEtlConfiguration anywhere in the patch. C# has both under ETL/Snowflake/.

src/Documents/Operations/Etl/Queue/AzureServiceBusConnectionSettings.ts
```diff
@@ -0,0 +1,81 @@
+import { StringUtil } from "../../../../Utility/StringUtil.js";
```
M4xymm, line 1: C# ETL/Queue/ also contains AmazonSqsConnectionSettings and AzureQueueStorageConnectionSettings; only Azure Service Bus was ported

src/Documents/Operations/AI/Agents/GetConversationMessagesOperation.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    async setResponseAsync(bodyStream: Stream, fromCache: boolean): Promise<string> {
+        // A missing conversation is a 404 with an empty body: the result stays null,
+        // whether the executor passes a null stream or the body carries no bytes.
+        if (!bodyStream) {
+            this.result = null;
+            return null;
+        }
+
+        const chunks: Buffer[] = [];
```
M4xymm: Manual stream buffering (chunks + finishedAsync + stringToReadable) reimplements what this._defaultPipeline(_ => body = _) already does in every other command in this repo. Prefer the shared idiom unless the empty-body-on-404 case is proven unreachable through it

M4xymm (changes requested):
Strong depth where it goes (full CdcSink Schema/ + Test/ surface, extra RunConversationOperation work, careful wire-fidelity comments), and everything compiles, lints, and passes export checks. However, there is one functional regression (dropping Disabled from the edit-certificate payload), the release plumbing is missing (no package.json/lock bump), and no Snowflake ETL, no Amazon SQS / Azure Queue Storage settings and no CdcSinkProcessState

redknightlois:
This is a new series that replaces the first one. These are the rules it was held to:

- Only the 7.2.3 to 7.2.5 changes of `src/Raven.Client` are in scope. Anything the .NET client already had at 7.2.3 that was never ported is a separate job, so it is neither added nor half added here.
- Visibility follows the .NET client: what it marks internal, or leaves out of the payloads it sends, stays out of the public API.
- When a request payload is touched, it is brought to parity: a field the .NET client always sends that this client drops is a bug in that file, however old.
- New code is shaped like the existing code next to it in this repo.
- `CLIENT_VERSION` tracks the server version; package.json is the npm release number and is bumped in the release commits.

How that applies to each review point:

- `Disabled` missing from the edit payload: fixed. `EditClientCertificateOperation` takes a `disabled` parameter (default false) and always sends `Disabled`; the SSO fields are sent only when given, and `Usage` is never sent. Covered by `CertificateSsoTest`.
- `disabled` missing from `CertificateMetadata`: not added in this series. The 7.2.3 to 7.2.5 change adds only the four SSO members to that class, and the read path converts the key casing of the whole response, so `disabled` already arrives at runtime; only the type does not name it. It is a one-line addition if you want it declared.
- `Snowflake` and `SnowflakeEtl` in the type unions: removed. `SnowflakeConnectionString` predates 7.2.3 and this client never had it, so no union member without a class behind it. One thing to know: `GetServerWideConnectionStringsOperation` maps each entry by its type, as the .NET client does, and throws on a type this client does not know, so a server-wide Snowflake connection string would hit that.
- `AmazonSqsConnectionSettings` and `AzureQueueStorageConnectionSettings`: not ported. Both already existed in the .NET client at 7.2.3; the 7.2.3 to 7.2.5 change only added Azure Service Bus. They are a separate port.
- Manual stream buffering in `GetConversationMessagesOperation`: replaced with the shared `this._defaultPipeline(_ => body = _)`, with a null check for the empty body on 404.
- package.json and package-lock bump: left as is, matching how this repo does it (the "Update version to 7.2.x in client" commits are separate); `CLIENT_VERSION` moves to 7.2.5.
- `CdcSinkProcessState`: added, and not exported, the same way `QueueSinkProcessState` is kept in this repo.
- Dropped compared with the first series: the CDC sink schema and mapping test operations (internal classes in the .NET client), the `cancelPendingActionTools` constructor plumbing on `AiConversation` (internal in the .NET client; the query parameter is sent as `false`, which is all a public caller can produce there), `CdcSinkTaskState` (a server-side state document with no client operation) and the `X-Forwarded-For` constant (nothing in the .NET client uses it).
- Tests: one file per feature under `test/Ported`, checking the request and response shapes where no server is needed.

M4xymm:
Tests are failing. 

`CdcSinkCrudTests`, `ServerWideConnectionStringsTest` tests should look like this:

```
((RavenTestContext.isRavenDbServerVersion("7.2") && !RavenTestContext.isPullRequest) ? describe : describe.skip)("CdcSinkCrudTest", function () {
```

Since the tests run during the PR's do not have a license, tests run manually on this branch will also verify the functionality of that pull request

src/Documents/Operations/AI/AiOperations.ts
```diff
@@ -74,4 +75,21 @@ export class AiOperations {
...
+    /**
+     * Reads messages from an AI conversation with full control over paging and filtering.
+     */
+    public async getConversationMessages(parameters: GetConversationMessagesOptions): Promise<AiConversationMessagesResult>;
+
+    public async getConversationMessages(conversationIdOrParameters: string | GetConversationMessagesOptions): Promise<AiConversationMessagesResult> {
+        const operation = typeof conversationIdOrParameters === "string"
+            ? new GetConversationMessagesOperation(conversationIdOrParameters)
+            : new GetConversationMessagesOperation(conversationIdOrParameters);
+        return await this._executor.send(operation);
```
M4xymm: This ternary operator returns exactly the same result

src/Documents/Operations/CdcSink/CdcColumnMapping.ts
```diff
@@ -0,0 +1,24 @@
...
+            column: this.column,
+            name: this.name
+        };
+
+        if (this.type && this.type !== "Default") {
+            json.type = this.type;
+        }
+
+        return json;
+    }
```
M4xymm: Do not migrate `toJSON` method

